PR #11, opened 2015-12-02
Global refactoring

pavel-kirienko:
This is still a work-in-progress.

@bendyer I'd like to ask a few questions concerning your implementation before I continue.

* [`_instantiate()` method](https://github.com/UAVCAN/pyuavcan/blob/master/uavcan/__init__.py#L93-L99). It is initialized in `load_dsdl()`, but it is not called from anywhere. Why do we need it?

* Field initialization via the constructor of `CompoundValue` seems to be intended, but not implemented. Is my understanding correct that the proper way of adding this feature would be adding necessary `setattr()` calls [here](https://github.com/UAVCAN/pyuavcan/blob/master/uavcan/transport.py#L376)? E.g.:

```python
for name, value in kwargs.items():
    setattr(self, name, value)
```

* The logic for handling incoming messages/requests/responses seems overengineered. I'm talking about `node.Monitor` and `node.Service`. I suggest to replace that with more Pythonic-looking callables, something along the lines of:

```python
def handle_service_request(request)
    response = process(request)
    return response

node.add_handler(MyServiceType, handle_service_request)
node.add_handler(MyMessageType, print)  # Print received messages
```

The proposed approach is also compatible with the existing interface-based approach.

---

It would be great if you could also review my other changes. They are not finished yet, but any feedback would be welcome. I am particularly interested in your opinion concerning the new `spin()` method instead of `listen()` (see my email from Nov 19):

```python
import uavcan
n = uavcan.make_node('/dev/ttyACM0', node_id=12, bitrate=100000, baudrate=115200)
n.periodic(1, lambda: n.send_message(uavcan.protocol.debug.KeyValue()))
n.spin(60)
print('Exit after 1 minute')
```

I had to temporarily drop the support for Tornado because it was getting in the way of refactoring. We can bring it back later if you still need it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,6 @@ target/
 
 # Mac stuff
 .DS_Store
+
+# Eclipse
+.settings/
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,6 @@ target/
 
 # Eclipse
 .settings/
+
+# IntelliJ
+.idea/
--- a/README.md
+++ b/README.md
@@ -14,7 +14,9 @@ UAVCAN is a lightweight protocol designed for reliable communication in aerospac
 
 ## Installation
 
-This library is dependency-free. Compatible Python versions are 2.7 and 3.3 and newer.
+Compatible Python versions are 2.7 and 3.3 and newer.
+If the library is used with Python 3, which is recommended, it does not require any additional dependencies.
+If Python 2.7 is used, additional dependencies are needed - refer to `setup.py` for more info.
 
 ```bash
 pip install uavcan
@@ -22,4 +24,5 @@ pip install uavcan
 
 ## Development
 
-The code should be formatted in compliance with [PEP8](https://www.python.org/dev/peps/pep-0008/).
+The code should be formatted in compliance with [PEP8](https://www.python.org/dev/peps/pep-0008/),
+with one exception: line length must not exceed 120 characters (PEP8 requires 79).
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import os
+import sys
 from setuptools import setup
 
 args = dict(
@@ -26,6 +27,9 @@ args = dict(
     ],
     keywords=''
 )
+
+if sys.version_info[0] < 3:
+    args['install_requires'] = ['monotonic']
 
 setup(**args)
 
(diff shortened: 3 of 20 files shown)


bendyer:
> `_instantiate()` method. It is initialized in `load_dsdl()`, but it is not called from anywhere. Why do we need it?

It's called from here:
https://github.com/UAVCAN/pyuavcan/blob/master/uavcan/dsdl/parser.py#L205-L206

We need it to support the current dynamic class constructors, e.g. `uavcan.protocol.NodeStatus()`.

pavel-kirienko:
@bendyer This is ready for review.

I have implemented field initialization via constructor:

```python
>>> uavcan.protocol.debug.KeyValue(value=1.23)
uavcan.protocol.debug.KeyValue(value=1.2300000190734863, key=ArrayValue(type=saturated uint8[<=58], tao=True, items=[]))
```

Library-specific fields such as `mode` and `tao` are now prefixed with underscore in order to avoid name clashing.

Request and response structures can be instantiated as follows:

```python
>>> uavcan.protocol.GetNodeInfo.Request()
uavcan.protocol.GetNodeInfo()

>>> uavcan.protocol.GetNodeInfo.Response()
uavcan.protocol.GetNodeInfo(status=uavcan.protocol.NodeStatus(uptime_sec=None, health=None, mode=None, sub_mode=None, vendor_specific_status_code=None), software_version=uavcan.protocol.SoftwareVersion(major=None, minor=None, optional_field_flags=None, vcs_commit=None, image_crc=None), hardware_version=uavcan.protocol.HardwareVersion(major=None, minor=None, unique_id=ArrayValue(type=saturated uint8[16], tao=False, items=[None, None, None, None, None, None, None, None, None, None, None, None, None, None, None, None]), certificate_of_authenticity=ArrayValue(type=saturated uint8[<=255], tao=False, items=[])), name=ArrayValue(type=saturated uint8[<=80], tao=True, items=[]))
```

Message and service handlers can be added and removed using `x = node.add_handler()`, `x.remove()`, and `node.remove_handlers()`:

```python
def server_callback(event):
    print(event.transfer)
    print(event.node)
    response = process(event.request)
    return response

node.add_handler(namespace.ServiceType, server_callback)
```

Existing class-based approach is also supported:

```python
n.add_handler(uavcan.protocol.NodeStatus, uavcan.monitors.NodeStatusMonitor)
```

I would appreciate if you could have a brief look at this and let me know if you find the changes acceptable. If everything looks right, will you be able to reimplement Tornado support?

PR #25, opened 2017-10-03
fix off-by-one error in union tag and array length bit size computation

jschall:


pavel-kirienko:
Thanks, @jschall! Nice catch. Could you please also fix the unit tests?

pavel-kirienko:
Hey @jschall, nudge nudge!

jschall:
I will try to get to this tomorrow.

mkeeter:
I think that the fix to `union_tag_bits` may be incorrect: a union with two elements only needs 1 bit to identify which one is selected, so it should be `ceil(log2(num_elements))` as before.

The `+ 1` is only necessary for dynamic arrays, e.g. an array with `<=64` elements actually has 65 possible sizes (0, 1, 2, ... 64).

pavel-kirienko:
> I think that the fix to `union_tag_bits` may be incorrect: a union with two elements only needs 1 bit to identify which one is selected, so it should be `ceil(log2(num_elements))` as before.

This is true.

> The `+ 1` is only necessary for dynamic arrays, e.g. an array with `<=64` elements actually has 65 possible sizes (0, 1, 2, ... 64).

Of course, static arrays do not have a length field at all.

pavel-kirienko:
@jschall

jschall:
Done.

Issue #4, opened 2015-07-28
Expand test suite


bendyer:
Needs further coverage, in particular:
- DSDL parser;
- Data type instantiation;
- Serialization and deserialization;
- Protocol conformance.

PR #27, opened 2017-11-27
Python 2.7 compatabilty for byte arrays

JamesStewy:
Explicitly convert byte arrays to `bytearray` type to allow item iteration in python 2.7.

### DIFF
--- a/uavcan/app/dynamic_node_id.py
+++ b/uavcan/app/dynamic_node_id.py
@@ -19,7 +19,7 @@ logger = getLogger(__name__)
 
 
 def _unique_id_to_string(uid):
-    return ' '.join(['%02X' % x for x in uid]) if uid else None
+    return ' '.join(['%02X' % x for x in bytearray(uid)]) if uid else None
 
 
 class CentralizedServer(object):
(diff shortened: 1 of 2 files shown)


pavel-kirienko:
Hi James. Were you able to manually test it against Python 3.x? Our unit tests are a bit flaky tbh.

JamesStewy:
Ah. I didn't realise. Only gave it a quick try in a python3 interpreter and it worked. Is there a suitable test program I could use to test? Maybe the gui tool?

pavel-kirienko:
> Only gave it a quick try in a python3 interpreter and it worked.

Ok, that's good enough for now, merging. Perhaps you could craft a new unit test that would validate your changes and submit it via another PR? (https://github.com/UAVCAN/pyuavcan/issues/4)

JamesStewy:
> craft a new unit

Not something I have done before. How would you go about testing the change in `file_server.py`.

pavel-kirienko:
The challenge of writing a solid unit test is to create an environment where the code can be tested as if it were functioning IRL. Normally you would construct a node instance, a file server instance, initialize everything, and run them. Then you would construct a second node instance and send some requests to the file server and make sure it responds correctly. It's a lot of work, indeed. In other projects I find myself writing unit tests most of the time, not the application code itself.

It's OK if you can't do that, but if you can, you'd render a considerable help to the project.

JamesStewy:
Ok, that would be too much for me to take on right now, but it sounds like something I can take on at a later date.

PR #30, opened 2018-03-28
use much faster crcmod library for crc

jschall:


pavel-kirienko:
I have a better proposition. Consider the result of profiling you shared yesterday:

```python
24527    0.108    0.000    7.077    0.000 /usr/local/lib/python3.5/dist-packages/uavcan/node.py:295(_recv_frame)
24527    0.224    0.000    6.634    0.000 /usr/local/lib/python3.5/dist-packages/uavcan/transport.py:752(from_frames)
24527    0.064    0.000    4.885    0.000 /usr/local/lib/python3.5/dist-packages/uavcan/dsdl/parser.py:304(get_data_type_signature)
24527    0.039    0.000    4.780    0.000 /usr/local/lib/python3.5/dist-packages/uavcan/dsdl/parser.py:293(get_dsdl_signature)
24527    0.036    0.000    4.352    0.000 /usr/local/lib/python3.5/dist-packages/uavcan/dsdl/signature.py:61(compute_signature)
24527    4.294    0.000    4.298    0.000 /usr/local/lib/python3.5/dist-packages/uavcan/dsdl/signature.py:39(add)
```

We can eliminate CRC computation completely, no extra dependencies necessary. The result of the method `get_dsdl_signature` is dependent only on the definition of its data type, which can't change while the program is running. So I propose to keep the current CRC computation algorithm unchanged, but cache the output of `get_dsdl_signature`; alternatively, we can compute the signature when the object is constructed and keep the result; in that case, `get_dsdl_signature` will just return the precomputed result.

A fair bit of computation is happening also here in `get_data_type_signature`: https://github.com/UAVCAN/pyuavcan/blob/0fd4122fe6e128aabd00d5ca88aba5cc82518c9c/uavcan/dsdl/parser.py#L304-L318
We could apply caching or precomputation here as well.

pavel-kirienko:
Closing in favor of https://github.com/UAVCAN/pyuavcan/pull/31

PR #31, opened 2018-03-30
cache return value of get_data_type_signature

jschall:


uavcan/dsdl/parser.py
```diff
@@ -217,6 +217,7 @@ def __init__(self, full_name, kind, source_file, default_dtid, source_text):
         self.default_dtid = default_dtid
         self.kind = kind
         self.source_text = source_text
+        self.data_type_signature = None
```
pavel-kirienko: Per PEP8, private attributes should be prefixed with an underscore: `_data_type_signature`

pavel-kirienko:
@jschall nudge

pavel-kirienko:
Splendid! What about this one though https://github.com/UAVCAN/pyuavcan/pull/31/files#diff-b87cc498613d10c86b2abc6ecce499c8R220

PR #33, opened 2018-05-16
slcan driver: increase receive buffer size in non-Mac OS

OXINARF:
As requested by @pavel-kirienko in https://github.com/ArduPilot/ardupilot/issues/8166#issuecomment-389222002, this patch increases the slcan driver receive buffer to one million in non-Mac OS.

As explained in that issue, the buffer is too small when the bus has a high number of messages.

**This was only tested in Ubuntu 16.04**

### DIFF
--- a/uavcan/driver/slcan.py
+++ b/uavcan/driver/slcan.py
@@ -47,7 +47,11 @@ except AttributeError:
 #
 # Constants and defaults
 #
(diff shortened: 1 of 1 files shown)


pavel-kirienko:
:+1: :tada: :man_dancing: :1st_place_medal:

PR #38, opened 2018-06-12
Add CAN-FD support to pyuavcan

thirtytwobits:
This is a preview as it is largely untested. Please comment but
keep on a branch until more confidence is built.

Also note that I am unaware of any CAN-FD slcan devices on the
market so this driver cannot be modified yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,6 @@ target/
 
 # IntelliJ
 .idea/
+
+# VSCode (We welcome our new Microsoft overlords!)
+.vscode
--- a/uavcan/__init__.py
+++ b/uavcan/__init__.py
@@ -19,6 +19,7 @@ import struct
 import pkg_resources
 import time
 from logging import getLogger
+from abc import ABCMeta
 
 try:
     # noinspection PyStatementEffect
@@ -45,6 +46,18 @@ In order to fix this problem, consider either option:
 class UAVCANException(Exception):
     pass
 
+class CanBus(object):
+    __metaclass__ = ABCMeta
+
+class CanBusFD(CanBus):
+    def __init__(self):
+        super(CanBusFD,self).__init__()
+        self.max_payload_len = 64
+
+class CanBus2(CanBus):
+    def __init__(self):
+        super(CanBus2,self).__init__()
+        self.max_payload_len = 8
 
 from .version import __version__
 import uavcan.node as node
--- a/uavcan/driver/socketcan.py
+++ b/uavcan/driver/socketcan.py
@@ -29,6 +29,8 @@ except ImportError:
 
 logger = getLogger(__name__)
 
+# CAN-FD flags for tx frames
+CANFD_FLAG_BRS = 1
 
 # Python 3.3+'s socket module has support for SocketCAN when running on Linux. Use that if possible.
 # noinspection PyBroadException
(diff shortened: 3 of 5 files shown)


.gitignore
```diff
@@ -68,3 +68,6 @@ target/
 
 # IntelliJ
 .idea/
+
+# VSCode (We welcome our new Microsoft overlords!)
```
pavel-kirienko: Out of curiosity, did you add this yourself, or was it added by Github automatically somehow?

uavcan/__init__.py
```diff
@@ -45,6 +46,18 @@ def monotonic_wrapper(*args, **kwargs):
 class UAVCANException(Exception):
     pass
 
+class CanBus(object):
```
pavel-kirienko: We seem to be leveraging a different naming convention where acronyms are all capitalized: `CANBus`, `CANBusFD`, `CANBus2`.

uavcan/driver/socketcan.py
```diff
@@ -29,19 +29,27 @@
 
 logger = getLogger(__name__)
 
+# CAN-FD flags for tx frames
+CANFD_FLAG_BRS = 1
 
 # Python 3.3+'s socket module has support for SocketCAN when running on Linux. Use that if possible.
```
pavel-kirienko: Is this not a perfect opportunity to drop support for Python 2.7? I was considering to do that with my next PR to this library, where I intend to *finally* add support for async API (the current callback-based API is a bit less friendly than it could be).

uavcan/node.py
```diff
@@ -29,7 +30,6 @@
 
 logger = getLogger(__name__)
 
-
```
pavel-kirienko: This extra line was [per PEP8](https://www.python.org/dev/peps/pep-0008/#blank-lines)

pavel-kirienko:
> Also note that I am unaware of any CAN-FD slcan devices on the market so this driver cannot be modified yet.

Yep, there aren't any, because SLCAN does not support CAN FD itself.

.gitignore
thirtytwobits: I added it. I'm using vs code these days. It's pretty great, actually.

kjetilkjeka:
Will these changes be tested in gui tool?

Is it much effort to make these changes available for gui_tool after the PR has landed?

pavel-kirienko:
> Is it much effort to make these changes available for gui_tool after the PR has landed?

The only thing requiring updates that comes to mind is the Bus Monitor widget - currently, it is built on the assumption that the maximum data length per frame is 8 bytes. Should be trivial to fix.

thirtytwobits:
I'm deep in the libuavcan tests today. I'll get to this PR next week. I'll also take a look at the gui_tool next week.

pavel-kirienko:
Could you please ask your guru if LPGL is okay? We can use LGPL bindings while keeping our codebase MIT-licensed.

scottdarch:
LPGL v3 is probably as bad if not worse. LGPL v2 is perhaps acceptable with review of the specific project and license.

pavel-kirienko:
PySide2 [is LGPLv3](https://groups.google.com/forum/#!topic/pyside-dev/pqwzngAGLWE).

I suppose purchasing the commercial license of PyQt5 is not an option either?

> I don't want to touch this repo until I get guidance from our OSS guru.

@thirtytwobits Pyuavcan is not affected by GPL, it is dependency-free.

scottdarch:
Sorry, by "this repository" I meant gui_tool which I was discussing how to test this change's effect in the gui_tool.

uavcan/driver/socketcan.py
thirtytwobits: Indeed. If we can drop 2.7 support this would be the way to go. It should offer some performance improvements as well since this API is more directly coupled with the Kernel.

pavel-kirienko:
Closing as obsolete.

PR #39, opened 2018-06-16
Also parse .uavcan files with explicit version number

kjetilkjeka:
I'm not really a python developer, hope i managed to not mess up these small changes to bad.

 - For the API part I assume that the correct thing to do was to add a version field to `CompondType`. My assumption is mainly based on `CompoundType` also containing the `default_dtid`, I've not done much research how the parser is actually used except for the `test.py` script in the dsdl repo.
 - I dont like the `_full_typename_version_and_dtid_from_filename` method name. Fortunately it's a private method, but perhaps `parse_filename` or something in that direction would be better?
 - I've verified that if two versions of the same message exists the list returned by `parse_namespace` will have two element with the same name. When they are printed they are indistinguishable, is that a problem?

### DIFF
--- a/test/test_transport.py
+++ b/test/test_transport.py
@@ -359,6 +359,7 @@ class TestArrayBasic(unittest.TestCase):
             parser.CompoundType.KIND_MESSAGE,
             "source.uavcan",
             0,
+            (0,0),
             ""
         )
         custom_type.fields = [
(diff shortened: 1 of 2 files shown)


kjetilkjeka:
Another thing i just realized. How do we handle version numbers in normalized definitions? I assume they should be added?

pavel-kirienko:
> How do we handle version numbers in normalized definitions? I assume they should be added?

Yes.

kjetilkjeka:
The point about normalized definitions is no longer valid as they will be removed with the protocol update. That's good :)

The point about the `_full_typename_version_and_dtid_from_filename` method name still stand. However, If a better name is found It should be easy enough to fix later.

---

I've been incredibly busy the last two weeks and will be for a couple more days. I will hopefully have a bit more time next week.

Issue #45, opened 2018-10-01
Exception in callbacks are caught by PyUAVCAN


antoinealb:
Sometimes I hit a bug or an exception in a callback. However this never hits my debugger because UAVCAN  catches them and ignores them.

Example: https://github.com/UAVCAN/pyuavcan/blob/master/uavcan/node.py#L217-L220

I would suggest passing a flag to the node or the spin method, something like `node.spin(catch_callback_exceptions=False)` to let advanced users use exceptions like they want.

What do you think? Maybe I am missing something on this design choice?

pavel-kirienko:
Pyuavcan does that to ensure that exceptions that occur in a handler do not prevent the library from invoking other handlers. Adding a flag as you suggested seems sensible. Perhaps it's time for a pull request?

antoinealb:
PR opened: #46

antoinealb:
Closed in #46

PR #46, opened 2018-10-04
Re-raise callback exceptions if asked to

antoinealb:
This commit fixes issue #45, by adding a flag to the spin method. It
allows the user to catch exceptions thrown from messages handler. This
is useful during debugging, so that thrown exceptions can be caught by
debuggers.

Not yet tested on a real project as I don't have hardware with me. Will report testing result ASAP.

### DIFF
--- a/uavcan/node.py
+++ b/uavcan/node.py
@@ -190,11 +190,7 @@ class HandlerDispatcher(object):
     def call_handlers(self, transfer):
         for uavcan_type, wrapper in self._handlers:
             if uavcan_type == get_uavcan_data_type(transfer.payload):
-                # noinspection PyBroadException
-                try:
-                    wrapper(transfer)
-                except Exception:
-                    logger.error('Transfer handler exception', exc_info=True)
+                wrapper(transfer)
 
 
 class TransferHookDispatcher(object):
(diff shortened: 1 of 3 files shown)


pavel-kirienko:
I noticed that having to carry this parameter around seems clumsy and redundant. Would it not be easier to supply it to the node constructor and make it an unchangeable private attribute?

antoinealb:
I changed the API to make this a node constructor parameter & added some unit tests.

uavcan/node.py
```diff
@@ -190,11 +190,7 @@ def remove_handlers(self, uavcan_type):
     def call_handlers(self, transfer):
         for uavcan_type, wrapper in self._handlers:
             if uavcan_type == get_uavcan_data_type(transfer.payload):
-                # noinspection PyBroadException
-                try:
-                    wrapper(transfer)
-                except Exception:
-                    logger.error('Transfer handler exception', exc_info=True)
+                wrapper(transfer)
```
pavel-kirienko: This change doesn't look correct. With your new logic, the first failing handler will prevent the library from invoking other handlers, which defeats the purpose of catching the exception in the library. Unless I misunderstood your intent, the old logic should be restored, with a try/except in every iteration.

antoinealb: Oh yeah I did not understand the original logic correctly. Pushed a fix.

pavel-kirienko:
Travis failed on Python 2.7 because of the exception handling syntax. Should we just ignore that, considering that the support for Python 2.7 is to be removed soon?

antoinealb:
I think it is just because python 2 is lacking `unittest.mock`, installing mock should fix the issue (pushed now).

pavel-kirienko:
I can't see your latest commit and a new CI build hasn't been triggered, are you sure it got pushed successfully?

antoinealb:
Sorry, I think I got distracted between sending my comment and pushing my code. It should be OK now.

PR #50, opened 2018-11-13
Add cache to compute_signature

NerdyProjects:
compute_signature creates a huge load on slow hardware (e.g. raspberry
pi). Using the functools lru cache removes most of that. Still, a higher
level cache that gets reused between instances of the same data type
would be more effective.
In practice, this commit is the difference between usable and unusable
for my raspberry pi logging some stuff with ~20 packets per second.

Big downside:
`functools.lru_cache` is Python 3.2+ only!

### DIFF
--- a/uavcan/dsdl/signature.py
+++ b/uavcan/dsdl/signature.py
@@ -8,6 +8,7 @@
 #
 
 from __future__ import division, absolute_import, print_function, unicode_literals
+import functools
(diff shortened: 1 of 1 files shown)


pavel-kirienko:
Can we not make a decorator wrapper that will do nothing if there's no `lru_cache` in functools?

PR #52, opened 2019-01-23
Travis CI: Drop Python 3.3 and add Python 3.6 and 3.7

cclauss:
Python 3.3 was end of life years ago and 3.4 end of life is in less than 60 days... https://devguide.python.org/#branchstatus

Also, [Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,14 @@
-sudo: false
 language: python
 python:
   - "2.7"
-  - "3.3"
   - "3.4"
   - "3.5"
+  - "3.6"
+matrix:
+  include:
+    - python: "3.7"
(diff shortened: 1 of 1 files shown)


pavel-kirienko:
Thanks.

cclauss:
Python 2 output wants __pip install monotonic__ added.

Python 3 output contains an exception.

pavel-kirienko:
> Python 2 output wants pip install monotonic added.

This dependency is optional and is not required.

> Python 3 output contains an exception.

I only see a test exception, which is thrown by design.

Issue #40, opened 2018-08-14
CAN acceptance filter


JamesStewy:
Hi Pavel,

I have a Raspberry Pi plugged into a Babel which is then connected into a CAN network that has a GNSS 2 and a Pixhawk. On the Raspberry Pi I am running a pyuavcan application and I am having a lot of performance issues if there is any non-trivial amount of traffic on the CAN network. Running `htop` shows that pyuavcan pegs one cpu core at 100%. With #33 now I am also getting a significant delay between the message being sent on the CAN network and it appearing in my pyuavcan application.

Can I suggest the addition of a CAN acceptance filter (or a blacklist) in pyuavcan that stops pyuavcan from processing certain messages and services? That way I can then filter out any high speed messages and save a lot of processing time. 

Thanks.

pavel-kirienko:
> Can I suggest the addition of a CAN acceptance filter (or a blacklist) in pyuavcan that stops pyuavcan from processing certain messages and services?

Eh. I guess we can do that, but it would only make sense doing that at the driver layer, immediately after the CAN frame is received from the adapter: https://github.com/UAVCAN/pyuavcan/blob/bfa4241a1a029f48a8683f214ee3087c4aa885be/uavcan/driver/slcan.py#L199

Getting rid of unwanted frames that early will also resolve the latency issue you mentioned. If you'd like to go ahead with that, feel free to submit a pull request.

But it won't solve your problems, really. Raspberry Pi is a calculator on steroids; if you need higher throughput and faster hardware is not an option, migrate to C++ and use libuavcan instead. Libuavcan supports automated acceptance filter configuration (courtesy @ilia-sheremet ) and hardware or software acceptance filters at the driver layer.

JamesStewy:
I have a minimal working version of what I need for my application [over here](https://github.com/MonashUAS/canros/tree/can_acceptance_filter). You just set `rx_blacklist` to a list of data type ids to be filtered.

To test it, I again used a GNSS 2 and then filtered out all messages it sends apart from its `NodeStatus` message. It works and the cpu usage is much lower now (at around 15% according to `htop`), but it isn't as low as I expected.

> But it won't solve your problems, really. Raspberry Pi is a calculator on steroids

I know the Pi isn't the fastest computer but I don't really understand why pyuavcan is much slower than libuavcan on a microcontroller.

pavel-kirienko:
[...]

PR #53, opened 2019-04-02
Add python can driver.

windelbouwman:


uavcan/driver/python_can.py
```diff
@@ -0,0 +1,157 @@
...
+
+        # get bus from default config
+        try:
+            self._bus = can.Bus()
+            self._bus.state = can.bus.BusState.ACTIVE
+        except Exception as ex:
+            logger.error("Got an exception", exc_info=True)
+
+        self._writer_thread_should_stop = False
+        self._write_queue = queue.Queue(self.TX_QUEUE_SIZE)
```
windelbouwman: Would it be okay to do without this queue?

windelbouwman:
@pavel-kirienko could you have a look at this added driver?

uavcan/driver/__init__.py
```diff
@@ -28,10 +29,13 @@ def make_driver(device_name, **kwargs):
...
     windows_com_port = device_name.replace('\\', '').replace('.', '').lower().startswith('com')
     unix_tty = device_name.startswith('/dev/')
+    pythoncan = device_name.startswith('pythoncan')
 
     if windows_com_port or unix_tty:
         return SLCAN(device_name, **kwargs)
     elif SocketCAN is not None:
         return SocketCAN(device_name, **kwargs)
+    elif pythoncan is not None:
+        return PythonCan(device_name, **kwargs)
```
pavel-kirienko: Move this condition one block higher, please. Otherwise, python_can would never be used on GNU/Linux.

uavcan/driver/python_can.py
```diff
@@ -0,0 +1,157 @@
...
+except ImportError:
+    # noinspection PyPep8Naming,PyUnresolvedReferences
+    import Queue as queue
+
+logger = getLogger(__name__)
+
+try:
+    import can
+except ImportError:
+    logger.error("no module 'can' available, please install python-can")
```
pavel-kirienko: This can't be an error. If the user doesn't care about Python-CAN, there is no use warning them about a library they don't even need being unavailable. I suggest demoting this to INFO.

uavcan/driver/python_can.py
```diff
@@ -0,0 +1,157 @@
...
+
+class PythonCan(AbstractDriver):
+    TX_QUEUE_SIZE = 1000
+
+    def __init__(self, interface, **_extras):
+        super(PythonCan, self).__init__()
+
+        # get bus from default config
+        try:
+            self._bus = can.Bus()
```
pavel-kirienko: This would fail in a very obscure way if the module couldn't be imported above. I suggest checking whether Python CAN is available early in the constructor, and raising a well-formed exception if it's not explaining what happened.

uavcan/driver/python_can.py
pavel-kirienko: As I understand, Python-CAN may be blocking, so I wouldn't remove the queue to keep the semantics of our various drivers compatible.

uavcan/driver/python_can.py
windelbouwman: Agreed, good idea.

uavcan/driver/__init__.py
pavel-kirienko: Any chance to have this one fixed?

uavcan/driver/python_can.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+class PythonCan(AbstractDriver):
+    TX_QUEUE_SIZE = 1000
+
+    def __init__(self, interface, **_extras):
+        super(PythonCan, self).__init__()
+
+        if can is None:
+            logger.error("To use this drive, make sure the module python-can is installed")
```
pavel-kirienko:
```suggestion
            logger.error("To use this driver, make sure the module python-can is installed")
```

uavcan/driver/python_can.py
```diff
@@ -0,0 +1,166 @@
...
+            raise ValueError("python-can must be installed.")
+
+        # get bus from default config
+        try:
+            self._bus = can.Bus()
+            self._bus.state = can.bus.BusState.ACTIVE
+        except Exception as ex:
+            logger.info("To use this drive, make sure the module python-can is installed")
```
pavel-kirienko: This seems unnecessary. If we're here, it's certainly installed.

uavcan/driver/python_can.py
```diff
@@ -0,0 +1,166 @@
...
+            logger.info("Also make sure you have a valid can.ini file located in the current working directory.")
+            logger.info("See also: https://python-can.readthedocs.io/en/master/configuration.html#configuration-file")
+            logger.error("Got an exception", exc_info=True)
```
pavel-kirienko:
```suggestion
            logger.exception("Could not instantiate a python-can driver")
```

adolfogc:
>**Filtering** 
The implementation features efficient filtering of can_id’s. That filtering occurs in the kernel and is much much more efficient than filtering messages in Python.

Source: https://python-can.readthedocs.io/en/master/interfaces/socketcan.html#filtering

Having this new driver available could enable a possible solution for #40?

windelbouwman:
I updated the code as you suggested.

uavcan/driver/python_can.py
```diff
@@ -0,0 +1,165 @@
...
+        super(PythonCan, self).__init__()
+
+        if can is None:
+            logger.error("To use this driver, make sure the module python-can is installed")
+            raise ValueError("python-can must be installed.")
+
+        # get bus from default config
+        try:
+            self._bus = can.Bus()
+            # self._bus.state = can.bus.BusState.ACTIVE
```
adolfogc: Nitpick: please remove commented code.

uavcan/driver/python_can.py
```diff
@@ -0,0 +1,165 @@
...
+    TX_QUEUE_SIZE = 1000
+
+    def __init__(self, interface, **_extras):
+        super(PythonCan, self).__init__()
+
+        if can is None:
+            logger.error("To use this driver, make sure the module python-can is installed")
+            raise ValueError("python-can must be installed.")
+
+        # get bus from default config
```
adolfogc: I suggest the following changes here:

```python
try:
    if interface is None:
        self._bus = can.Bus() # get bus from environment's config
    else:
        self._bus = can.interface.Bus(channel=interface, bustype=_extras['bustype'], bitrate=_extras['bitrate'])
except Exception as ex:
    # ...
```

This allows backward compatibility by maintaining the interface used by the other drivers.

pavel-kirienko:
Hey @windelbouwman, could you please address the outlined issues so that we could merge this PR? Thanks!

uavcan/driver/__init__.py
windelbouwman: Done.

uavcan/driver/python_can.py
windelbouwman: Done.

windelbouwman:
@pavel-kirienko I think this change is ready now. I addressed the issues.

uavcan/driver/__init__.py
```diff
@@ -28,9 +29,12 @@ def make_driver(device_name, **kwargs):
     """
     windows_com_port = device_name.replace('\\', '').replace('.', '').lower().startswith('com')
     unix_tty = device_name.startswith('/dev/')
+    pythoncan = device_name.startswith('pythoncan')
 
     if windows_com_port or unix_tty:
         return SLCAN(device_name, **kwargs)
+    elif pythoncan is not None:
```
pavel-kirienko: Have you tested this code? This condition will never be true because `pythoncan` is of type `bool`, hence can't be None.

uavcan/driver/python_can.py
pavel-kirienko: @windelbouwman this is a valid suggestion, please address.

uavcan/driver/__init__.py
windelbouwman: No, I did not test this code. I now tested it. The code is somewhat simpler, but works now. No option to set the bitrate via the `kwargs` parameter. In anyway, we set it in the can.ini file.

windelbouwman:
I implemented the review comments, please check if this is sufficient.

uavcan/driver/python_can.py
```diff
@@ -34,20 +34,15 @@
...
         if can is None:
             logger.error("To use this driver, make sure the module python-can is installed")
             raise ValueError("python-can must be installed.")
 
-        # get bus from default config
         try:
-            if interface is None:
-                self._bus = can.Bus()
-            else:
-                self._bus = can.interface.Bus(channel=interface, bustype=_extras['bustype'], bitrate=_extras['bitrate'])
```
adolfogc: @windelbouwman Why delete this?

windelbouwman: The interface in this case will be `pythoncan`, which is an invalid channel. We choose `pythoncan` as interface, so it makes no sense to pass it. For the second and third argument, I was unable to make it work, and I also do not use it yet. In any case, it is not required to check if interface is `None`.

adolfogc: Thanks for your reply; `interface` was a variable that referred to the underlying CAN interface, e.g., can0 or vcan0, not to `pythoncan`. Regarding the other parameters, they are needed as well when not using an INI file.

> I would like to merge this driver first as it is, we can later refine it?

OK, no problem, I'll submit a PR with these changes afterwards.

PR #66, opened 2019-08-01
Improve python-can driver

adolfogc:
- Fixes the issue of not having python-can installed causing an error, even though users may want to use, e.g., SocketCAN instead.
- Allows passing CAN interface configuration information using the constructor instead of requiring the use of an INI file.
- Corrects other minor issues.

### DIFF
--- a/uavcan/driver/__init__.py
+++ b/uavcan/driver/__init__.py
@@ -11,12 +11,18 @@ from __future__ import division, absolute_import, print_function, unicode_litera
 import sys
 from .slcan import SLCAN
 from .common import DriverError, CANFrame
-from .python_can import PythonCan
 
 if sys.platform.startswith('linux'):
     from .socketcan import SocketCAN
 else:
     SocketCAN = None
+
+try:
+    import can
+except ImportError:
+    PythonCAN = None
+else:
+    from .python_can import PythonCAN
 
 __all__ = ['make_driver', 'DriverError', 'CANFrame']
 
@@ -29,12 +35,11 @@ def make_driver(device_name, **kwargs):
     """
     windows_com_port = device_name.replace('\\', '').replace('.', '').lower().startswith('com')
     unix_tty = device_name.startswith('/dev/')
-    pythoncan = device_name.startswith('pythoncan')
 
     if windows_com_port or unix_tty:
         return SLCAN(device_name, **kwargs)
-    elif pythoncan:
-        return PythonCan()
+    elif PythonCAN is not None:
+        return PythonCAN(device_name, **kwargs)
     elif SocketCAN is not None:
         return SocketCAN(device_name, **kwargs)
     else:
--- a/uavcan/driver/__init__.py
+++ b/uavcan/driver/__init__.py
@@ -19,10 +19,9 @@ else:
 
 try:
     import can
+    from .python_can import PythonCAN
 except ImportError:
     PythonCAN = None
-else:
-    from .python_can import PythonCAN
 
 __all__ = ['make_driver', 'DriverError', 'CANFrame']
 
--- a/uavcan/driver/__init__.py
+++ b/uavcan/driver/__init__.py
@@ -9,6 +9,7 @@
 
 from __future__ import division, absolute_import, print_function, unicode_literals
 import sys
+from .python_can import PythonCAN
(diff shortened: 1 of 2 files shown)


adolfogc:
@pavel-kirienko In https://github.com/UAVCAN/pyuavcan/pull/53#pullrequestreview-241946587 you said:

> ... here is still a room for improvement (such as writing tests, documentation, and fixing minor mistakes such as the fixed latency being only 1 microsecond which is unrealistic). ...

I'm guessing the correct value depends on the hardware we're using, but I'm don't know how to get the correct one to use. What is your suggestion to address this issue?

uavcan/driver/__init__.py
```diff
@@ -11,13 +11,19 @@
...
 from .common import DriverError, CANFrame
-from .python_can import PythonCan
 
 if sys.platform.startswith('linux'):
     from .socketcan import SocketCAN
 else:
     SocketCAN = None
 
+try:
+    import can
```
pavel-kirienko: Why can't we move `from .python_can import PythonCAN` here?

thirtytwobits:
Why are the AppVeyor builds failing for master?

thirtytwobits:
OH! I just realized this should enable uavcan tools to run on OSX with the peak systems adapter. Yes please! I can haz this PR @pavel-kirienko ?

pavel-kirienko:
> Why are the AppVeyor builds failing for master?

Because there is no appveyor.yml on master. Could someone submit one, please?

> OH! I just realized this should enable uavcan tools to run on OSX with the peak systems adapter. Yes please! I can haz this PR @pavel-kirienko ?

Observe that this is UAVCAN v0 though. We could really use it in UAVCAN v1.

thirtytwobits:
@adolfogc : if you rebase this on mainline I've merged in an appveyor.yml that will fix the checks.

uavcan/driver/__init__.py
adolfogc: IMO it reads more explicit that we're just checking that the specific module, `can`, is present; also that way we also prevent a possible import error in `python_can.py` from being caught and ignored.

adolfogc:
> @adolfogc : if you rebase this on mainline I've merged in an appveyor.yml that will fix the checks.

Thanks @thirtytwobits

adolfogc:
> We should specify the best-case fixed latency attainable with common high-level operating systems such as Windows. My experience indicates that the value of 100~200 microseconds fits the majority of deployments. If you are interested in the underlying technical details, see the Olson passive time synchronization (recovery) algorithm: https://april.eecs.umich.edu/pdfs/olson2010.pdf

Thanks, @pavel-kirienko, I'll check that out.

> Would you be available to implement PythonCAN support for the new PyUAVCAN rewrite? It's in the branch uavcan-v1.0 and it's totally new (no old code whatsoever).

Yes, I intend to continue working on the v1 branch as well. This last month I've been focused on a course I'm attending abroad and the next one I'll be somewhat distracted because I'm moving overseas to study, so I expect not being able to work at a constant rate shortly. Is there a deadline I should keep in mind?

pavel-kirienko:
> Is there a deadline I should keep in mind?

There is no strict deadline but it would be optimal to have it done by September so that we could begin to push people towards v1 more confidently. Perhaps we should replace master with v1 around that time, too, but that is also dependent on @thirtytwobits' ongoing work on libuavcan.

uavcan/driver/__init__.py
pavel-kirienko: IDK it seems non-obvious and either I'm reading too much into it but it also looks like a leaky abstraction where you make assumptions about the underlying structure of the imported module. The general lack of a well-constructed architecture here might make my observation about the leaky abstraction here a bit inappropriate but it's still valid.

If you are certain about this then I won't insist.

adolfogc: I've made the change as per your request.

pavel-kirienko: That's not quite what I meant. The `import can` statement is the one I am concerned about. Any chance to get rid of it?

adolfogc: Sorry, I misunderstood what you meant. I've moved the check to the `.python_can` module itself. You're right, it's cleaner that way.

PR #68, opened 2019-08-08
Allowing duplicate definitions when parsing dsdl

thirtytwobits:
duplicate == the same thing more than once. This is okay.
redefined == the same identifer used by incompatible definitions.
this is not okay.

In complex build systems duplicate dsdl files can get pulled in from
multiple dependencies and end up in long lists of paths that are
passed into the pyuavcan dsdl compiler. This change to the dsdl
parser allows this without throwing errors but enforces that the
duplicated definitions are compatible. This elminiates the need
for special case handling in build systems to try to find the duplicates
before invoking pyuavcan and adds a new check where such complex
systems will be notified when duplicated definitions get out of
sync.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,10 @@ target/
 
 # IntelliJ
 .idea/
+
+# VS code
+.vscode
+
+# virtualenv
+.pyenv
+
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,12 @@
+environment:
+  matrix:
+    - APPVEYOR_BUILD_WORKER_IMAGE: Ubuntu
+      PYTHON: "3.7"
+
+stack: python %PYTHON%
+
+build: off
+branches:
+  except:
+    - gh-pages
+    - master
--- a/test/dsdl/__init__.py
+++ b/test/dsdl/__init__.py
@@ -1,0 +1,8 @@
+#
+# Copyright (C) 2014-2015  UAVCAN Development Team  <uavcan.org>
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# This software is distributed under the terms of the MIT License.
+#
+# Define dsdl folder as module so python unittest autodiscover will
+# work correctly
--- a/test/dsdl/fake_dsdl/ns0_base/ns0/20000.Type0.uavcan
+++ b/test/dsdl/fake_dsdl/ns0_base/ns0/20000.Type0.uavcan
@@ -1,0 +1,4 @@
+#
+# Base type for testing duplicate handling.
+#
+uint8 field0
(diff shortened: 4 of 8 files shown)


pavel-kirienko:
This does not quite contradict the v0 spec, but it would be against the v1 spec unless I misunderstood the described use case. We can accept this in v0 but not v1 unless the reasoning presented in this discussion is shown to be flawed (it is related to static compatibility checking we have defined for v1): https://forum.uavcan.org/t/new-dsdl-directive-for-simple-compile-time-checks/190/16?u=pavel.kirienko

It starts with this quote:

> PyDSDL prohibits the same root namespace from being defined in several different locations. Meaning that spreading definitions belonging to the same root namespace across different directories is not permitted (e.g. `/foo/bar/uavcan` , `/baz/bar/uavcan` ).

And several posts later ends here:

> What is prohibited is that you can’t invoke a parser with two root namespace directories both ending with the same final path component: `/foo/bar/uavcan` , `/abc/baz/uavcan` ; because that would imply that either the same definitions are supplied twice, or members of the same root namespace are distributed across several directories. The latter is dangerous because if a parser is ever invoked on only one of those directories, it will be unable to detect possible conflicts with the other.

Documented in the specification, section 3.1.3 File hierarchy:

> A directory defining a namespace will always define said namespace in its entirety, meaning that the contents of a namespace cannot be spread across different directories sharing the same name.

Is it true that you are proposing a mere v0 implementation detail and not suggesting to change that part of v1 spec?

thirtytwobits:
> Is it true that you are proposing a mere v0 implementation detail and not suggesting to change that part of v1 spec?

I guess I'm proposing that we allow this in v0 and discuss it for v1. I think we should only restrict what the types resolve to not where they come from. As I mentioned in my commit, this emerged for us as we deployed UAVCAN to distinct parts of a system and then recombined these parts in integration builds. Not only does allowing the duplication simplify the ability to deal with distributed development pipelines it actually provides a really useful compile-time check that two parts of a given system are still utilizing a compatible message definition when they are integrated. For V1 this becomes more sophisticated (read: the duplicate versus redefinition logic may be more involved) but the value is the same.

thirtytwobits:
Also: If you accept this in v0 I'll drive the changes in v1.

pavel-kirienko:
> I'll drive the changes in v1.

Whatever. I don't immediately see how the obvious issues with definition consistency and enforcing versioning constraints are resolved in the workflow case you described but if you think they are resolvable then sure let's discuss that (not here). When (if) starting that discussion, could you please also provide a practical example (real-life or constructed).

I should note that the underlying idea of keeping all members of a namespace in the same directory is to make types that coexist in the same environment at runtime (which is the UAVCAN bus they are used within) also coexist in the same environment at compile time (which is the root namespace directory). If you look at it this way, you'd probably see that centralized management of a namespace is desirable for ensuring wire compatibility.

PR #81, opened 2019-10-09
Fix Collections Deprecation Warning

ImXron:
Minor change to get rid of a deprecation warning in 3.7

```
DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```

Ref: https://docs.python.org/3/whatsnew/3.7.html#id3

### DIFF
--- a/uavcan/transport.py
+++ b/uavcan/transport.py
@@ -14,7 +14,11 @@ import math
 import copy
 import struct
 import functools
-import collections.abc
+
+try:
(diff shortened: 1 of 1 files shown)


ImXron:
Looking into the python 2 build fail.

pavel-kirienko:
This is good, thank you Imran.

Issue #95, opened 2020-02-19
Tests fail with PyTest v5.x


pavel-kirienko:
I tried upgrading to PyTest v5.3, but Nunavut stopped working when invoked by the testing suite:

```
NotImplementedError("Can't perform this operation for unregistered loader type")
Traceback (most recent call last):

  <...irrelevant...>

  File "/home/travis/build/UAVCAN/pyuavcan/pyuavcan/dsdl/_compiler.py", line 197, in generate_package
    namespace_output_stem='__init__')

  File "/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/nunavut/lang/__init__.py", line 225, in __init__
    self._config = self._load_config(*additional_config_files)

  File "/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/nunavut/lang/__init__.py", line 208, in _load_config
    resources = [r for r in pkg_resources.resource_listdir(__name__, '.') if r.endswith('.ini')]

  File "/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1164, in resource_listdir
    resource_name

  File "/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1442, in resource_listdir
    return self._listdir(self._fn(self.module_path, resource_name))

  File "/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1484, in _listdir
    "Can't perform this operation for unregistered loader type"

NotImplementedError: Can't perform this operation for unregistered loader type
```

It's super weird that the exception comes from `pkg_resources`. It does not seem to be involved here at all.

Nunavut works fine outside of PyTest. I googled around a bit but could not find a quick solution, so I decided to downgrade back to PyTest v4.6 for now.

thirtytwobits:
Ack. Yeah, I use pkg_resources to load templates from the package. I should probably switch to importlib.

jxltom:
Sorry update the I forgot to remove the closing keyword in the PR. This issue is not solved actually, please reopen  this. Sorry for the inconvenience.

pavel-kirienko:
Indeed, I didn't notice that it got closed automatically. Thanks for raising this.

pavel-kirienko:
@thirtytwobits should we transfer this issue over to Nunavut?

pavel-kirienko:
[...]

Issue #103, opened 2020-03-19
Support PyDSDL v1.3


pavel-kirienko:
https://github.com/UAVCAN/pydsdl/pull/34

pavel-kirienko:
https://github.com/UAVCAN/pydsdl/pull/39

PR #111, opened 2020-04-28
Update pydsdl as requirement to 1.4

jxltom:
Since https://github.com/UAVCAN/pyuavcan/issues/103 is solved by https://github.com/UAVCAN/pydsdl/pull/39, ```pydsdl``` as requirement can be locked to ```1.4```.

Dependency nunavut also requires pydsdl 1.4, this PR solves the conflict.

This PR also fixes tests in pytest 5.4, closes https://github.com/UAVCAN/pyuavcan/issues/95

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@
 
 # Testing.
 pytest          ~= 4.6
-pytest-asyncio  >= 0.10, < 2.0
+pytest-asyncio  == 0.10
 coverage        ~= 5.0
 
 # Static analysis. PyLint is needed for the SonarQube scanner.
--- a/setup.cfg
+++ b/setup.cfg
@@ -60,7 +60,7 @@ packages             = find:
 # Think thrice before adding anything here, please.
 # The preferred long-term plan is to avoid adding any new dependencies whatsoever for the project's lifetime.
 install_requires =
-    pydsdl  == 1.2.0 # TODO: issue #103 support new TaggedUnionType introduced in pydsdl 1.3
+    pydsdl  ~= 1.4
     numpy   ~= 1.16
     nunavut ~= 0.2
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -62,7 +62,7 @@ packages             = find:
 install_requires =
     pydsdl  ~= 1.4
     numpy   ~= 1.16
-    nunavut ~= 0.2
+    nunavut == 0.2.3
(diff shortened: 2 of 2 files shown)


pavel-kirienko:
Please also freeze `pytest-asyncio == 0.10` to fix the test suite. It broke because the recently released version is incompatible with PyTest v4, and we can't upgrade it because of https://github.com/UAVCAN/pyuavcan/issues/95.

A better solution would be to replace pkg_resources with importlib in Nunavut and then upgrade PyTest here to the latest version. Maybe you could do that instead?

jxltom:
The tests are fixed, it looks like including all python files as tests source by ```python_files     = *.py``` trigger the issue.

pavel-kirienko:
This is certainly not the root cause because the filename should have nothing to do with it. Your fix silently removes unit tests that are located in files named with a leading underscore so this is undesirable.

jxltom:
There are still 126 tests left, no tests are removed by this PR. It is better to name test files with prefix ```test_``` in the further development.

My suspect is that including all files in tests might also includes ```setup.py``` or whatever which is not tests at all trigger the issue, but I have no time to dig it out

pavel-kirienko:
`test_*.py` won't work because some tests are embedded directly in the implementation files to enhance test locality. This is an intentional design decision and I currently see no reason to change it. The underscore prefix would break the tests in the bootloader branch. The correct fix for the above problem is to fix Nunavut.

jxltom:
OK, I don't know there are embbeded tests in source dir in other branch.

I will check your suggestions.

arrowcircle:
This fix is also required for yukon, because it depends on nuvanut (needs 1.4.0) and pyuavcan (needs 1.2.0). Without it it's not possible to manage dependencies with pip.

jxltom:
Just have some time to continue the PR. Yeah you are right. The ```pkg_resources``` is not well supported since ```pytest 4.6.0``` as suggested here https://github.com/pytest-dev/pytest/issues/5392 and https://github.com/pytest-dev/pytest/commit/13f02af97d676bdf7143aa1834c8898fbf753d87. It is better to replace ```pkg_resources``` with ```importlib_resources``` and ```importlib-metadata```

However, I'm not sure when I have time to do ths, so in this PR, I just pin the pytest as well as its related requirements to ```v4.6``` temporarily.

PR #121, opened 2020-07-30
[legacy-v0] transport.py: Fix endianness of dynamic array lengths

pcarranzav2:
Dynamic array lengths are encoded as any other scalar, so when they're decoded/encoded we should
make sure they're read/written as little endian scalars. This change fixes the fact that converting binary
strings to/from ints assumes the string is in big endian representation, so arrays with max length over 256 would be incorrectly encoded.


pcarranzav2:
@pavel-kirienko thanks for the lightning-fast response!

pavel-kirienko:
yeah this is rather unusual for me

PR #123, opened 2020-08-13
python-can support implemented

Polarisru:
Tested with PCAN-USB Pro and Kvaser Leaf Light v2, some problems with Kvaser detected.


pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -16,42 +24,174 @@
 class PythonCANMedia(_media.Media):
     """
     A media interface adapter for `python-can <https://github.com/hardbyte/python-can>`_.
-    This is a stub, the implementation is missing. Please submit patches!
+      * Usage example for PCAN-USB channel 1 (bitrate = 500k, mtu = 8, Node-ID = 10): 
+          --tr="CAN(can.media.pythoncan.PythonCANMedia('pcan','PCAN_USBBUS1',5000000,8),10)"
+      * Usage example for Kvaser channel 0 (bitrate = 500k, mtu = 8, Node-ID = 10): 
+          --tr="CAN(can.media.pythoncan.PythonCANMedia('kvaser','0',5000000,8),10)"
```
pavel-kirienko, line 30: That's not quite how you write reStructuredText. This is how you do it so that it's rendered correctly:

```suggestion

    - Usage example for PCAN-USB channel 1 (bitrate = 500k, mtu = 8, Node-ID = 10)::

        --tr="CAN(can.media.pythoncan.PythonCANMedia('pcan','PCAN_USBBUS1',5000000,8),10)"

    - Usage example for Kvaser channel 0 (bitrate = 500k, mtu = 8, Node-ID = 10)::

        --tr="CAN(can.media.pythoncan.PythonCANMedia('kvaser','0',5000000,8),10)"
```

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -16,42 +24,174 @@
...
     @property
     def mtu(self) -> int:
-        raise NotImplementedError
+        return self._mtu
 
     @property
     def number_of_acceptance_filters(self) -> int:
-        raise NotImplementedError
+        #just a placeholder to avoid error
+        return 16
```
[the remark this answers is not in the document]
Polarisru: I don't know how to get the number of filters, but this value have to be defined, so I had to put a number there. How would be better to solve such an issue?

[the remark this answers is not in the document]
pavel-kirienko: It's better to err towards a smaller number then. Like 1.

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -16,42 +24,174 @@
...
+        for f in configuration:
+            f_dict = {}
+            f_dict['can_id'] = f.identifier
+            f_dict['can_mask'] = f.mask
+            f_dict['extended'] = f.format == _media.FrameFormat.EXTENDED
+            filters.append(f_dict)
+        _logger.info('Acceptance filters activated: %s', ', '.join(map(str, configuration)))
+        self._bus.set_filters(filters)
+                        
+    def send_msg(self, msg: can.Message) -> None:
```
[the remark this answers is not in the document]
Polarisru: Is a private method defined with double underscore? I see that private methods of _socketcan.py are defined with single underscore (_parse_native_frame for example). Is it a correct way here?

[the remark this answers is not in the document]
pavel-kirienko: It takes just one underscore: `_send_msg`, although in this specific case, if you also consider my comment about timeout, it's better to make a closure or just use a lambda, like, instead of:

```
await self._loop.run_in_executor(self._background_executor, self.send_msg, message)
```

You say something along the lines of:

```
await self._loop.run_in_executor(self._background_executor, lambda: self._bus.send(message, timeout=monotonic_deadline - self._loop.time()))
```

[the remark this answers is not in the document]
Polarisru: Ok, I see, seems to be a better solution here

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -16,42 +24,174 @@
...
-    This is a stub, the implementation is missing. Please submit patches!
+      * Usage example for PCAN-USB channel 1 (bitrate = 500k, mtu = 8, Node-ID = 10): 
+          --tr="CAN(can.media.pythoncan.PythonCANMedia('pcan','PCAN_USBBUS1',5000000,8),10)"
+      * Usage example for Kvaser channel 0 (bitrate = 500k, mtu = 8, Node-ID = 10): 
+          --tr="CAN(can.media.pythoncan.PythonCANMedia('kvaser','0',5000000,8),10)"
     """
 
-    def __init__(self) -> None:
-        raise NotImplementedError
+    def __init__(self, iface_name: str, channel_name: str, bitrate: int, mtu: int) -> None:
```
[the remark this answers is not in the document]
Polarisru: At the moment second bitrate is the same, but we can change it, python-can supports some FD features, so it should be possible to add data bitrate there.

I don't think it's a good idea to limit CAN FD packet with 64 bytes only, we have to add all possible packet sizes here.

May be could concatenate interface name and channel name as a ':'-separated string and process it internally?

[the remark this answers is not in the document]
pavel-kirienko: > I don't think it's a good idea to limit CAN FD packet with 64 bytes only, we have to add all possible packet sizes here.

I am not saying we should do that. I am saying that most applications will be using 64 bytes, so it makes sense to make it a default, so that the user won't have to type the `,8` at the end in `can.media.pythoncan.PythonCANMedia('pcan','PCAN_USBBUS1',5000000,8)`.

> May be could concatenate interface name and channel name as a ':'-separated string and process it internally?

Seems sensible. Observe that we will need to return the same name (concatenated with `:`) from the property method `iface_name`.

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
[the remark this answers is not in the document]
Polarisru: Ok, I will put 1 there.

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -16,42 +24,174 @@
...
+                                           format=frame_format,
+                                           loopback=loopback,
+                                           timestamp=timestamp)                
+
+    def _set_loopback_enabled(self, enable: bool) -> None:
+        # it do nothing at the moment
+        if enable != self._loopback_enabled:
+            self._loopback_enabled = enable
+        
+class _NativeFrameDataCapacity(enum.IntEnum):
```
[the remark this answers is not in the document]
Polarisru: I have removed this definition and will use self.VALID_MTU_SET now

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -16,42 +24,174 @@
...
+                if out is not None:
+                    return out
+                
+    @staticmethod
+    def _parse_native_frame(msg: can.Message,
+                            loopback: bool,
+                            timestamp: pyuavcan.transport.Timestamp) \
+            -> typing.Optional[_media.TimestampedDataFrame]:
+        if (msg.error_state_indicator == True) or (msg.is_error_frame == True):  # error frame, ignore silently
+            _logger.debug('Frame dropped: id_raw=%08x', ident_raw)
```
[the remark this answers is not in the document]
Polarisru: It will be fixed

pavel-kirienko:
Hey @Polarisru, is it ready for review? Do you need any assistance with this? Thanks!

Polarisru:
Hello, yes, I have committed the last version a week ago, can you see it? I have fixed probably all of your comments, but python-can has some problems regarding CAN FD implementation, the settings are very device-specific for different adapters, so I have just added CAN FD support for adapters I have only (PCAN and Kvaser).

pavel-kirienko:
Thanks. Yes, I've seen your commits but you didn't request another review so I thought that you are still working on it. I will give it a look soon.

pavel-kirienko:
Meanwhile, can you please fix the typing issues detected by MyPy? You can find them listed at https://ci.appveyor.com/project/Zubax/pyuavcan/builds/34717452/job/ryxadmg0fxnesff7 if you scroll to the very end:

```
tests/cli/_subprocess.py: note: In member "interrupt" of class "BackgroundChildProcess":
tests/cli/_subprocess.py:123: error: Module has no attribute "CTRL_BREAK_EVENT"
pyuavcan/transport/can/media/pythoncan/_pythoncan.py: note: In member "__init__" of class "PythonCANMedia":
pyuavcan/transport/can/media/pythoncan/_pythoncan.py:53: error: Value of type "Union[int, Tuple[int, int]]" is not indexable
pyuavcan/transport/can/media/pythoncan/_pythoncan.py:54: error: Value of type "Union[int, Tuple[int, int]]" is not indexable
pyuavcan/transport/can/media/pythoncan/_pythoncan.py:56: error: Argument 1 to "int" has incompatible type "Union[int, Tuple[int, int]]"; expected "Union[str, bytes, SupportsInt, _SupportsIndex]"
pyuavcan/transport/can/media/pythoncan/_pythoncan.py:57: error: Argument 1 to "int" has incompatible type "Union[int, Tuple[int, int]]"; expected "Union[str, bytes, SupportsInt, _SupportsIndex]"
pyuavcan/transport/can/media/pythoncan/_pythoncan.py: note: In member "send_until" of class "PythonCANMedia":
pyuavcan/transport/can/media/pythoncan/_pythoncan.py:144: error: Returning Any from function declared to return "object"
Found 6 errors in 2 files (checked 364 source files)
```

Also, at least some rudimentary test suite would help.

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -90,25 +129,23 @@ def configure_acceptance_filters(self, configuration: typing.Sequence[_media.Fil
...
                 raise pyuavcan.transport.ResourceClosedError(repr(self))
             self._set_loopback_enabled(f.loopback)
-            # check settings for CAN FD, not implemented yet!
             message = can.Message(arbitration_id=f.identifier, is_extended_id=True, data=f.data, is_fd=self._is_fd)
             if f.loopback == True:
                 with self._loopback_lock:
                     self._loop_frames.append(f)
             try:
-                await self._loop.run_in_executor(self._background_executor, self.send_msg, message)
+                await self._loop.run_in_executor(self._background_executor, lambda: self._bus.send(message, timeout=monotonic_deadline - self._loop.time()))
```
pavel-kirienko, line 144: The line is too long

```suggestion
                await self._loop.run_in_executor(
                    self._background_executor,
                    lambda: self._bus.send(message, timeout=monotonic_deadline - self._loop.time()),
                )
```

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -174,22 +211,22 @@ def _read_frame(self) -> _media.TimestampedDataFrame:
...
                 if out is not None:
                     return out
-                
+
     @staticmethod
     def _parse_native_frame(msg: can.Message,
                             loopback: bool,
                             timestamp: pyuavcan.transport.Timestamp) \
             -> typing.Optional[_media.TimestampedDataFrame]:
         if (msg.error_state_indicator == True) or (msg.is_error_frame == True):  # error frame, ignore silently
```
pavel-kirienko, line 220: If I'm reading the PythonCAN docs correctly, [`error_state_indicator`](https://python-can.readthedocs.io/en/master/message.html#can.Message.error_state_indicator) merely indicates that the adapter is in the error active state, so it should not be checked here.

```suggestion
        if msg.is_error_frame:  # error frame, ignore silently
```

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -26,25 +26,64 @@ class PythonCANMedia(_media.Media):
...
 
-    def __init__(self, iface_name: str, channel_name: str, bitrate: int, mtu: int) -> None:
+    VALID_FD_ADAPTER_SET = {'pcan', 'kvaser'}
+
+    def __init__(self, iface_name: str, bitrate: typing.Union[int, typing.Tuple[int, int]], mtu: int) -> None:
         self._iface_name = str(iface_name)
-        self._channel_name = str(channel_name)
+        self._conn_name = self._iface_name.split(':')
+        if len(self._conn_name) != 2:
+            raise RuntimeError('Wrong interface name format')
```
pavel-kirienko, line 42: Please do not use generic error types because proper error handling is part of the API contract. The user may be looking to catch specific types to handle them in a particular way; this is why we have a hierarchy of exception types (described in PyUAVCAN docs). It is also always better to provide more context in the error message to help with troubleshooting.

```suggestion
            raise pyuavcan.transport.InvalidMediaConfigurationError(
                "Interface name %r does not match the format 'interface:channel'" % self._iface_name
            )
```

Speaking of context. It's a pet peeve of mine -- I don't understand the modern tendency in UI/UX design to strip error messages from useful information. This:

image

Is a huge step back from this:

image

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -26,25 +26,64 @@ class PythonCANMedia(_media.Media):
...
+        if mtu not in self.VALID_MTU_SET:
+            raise RuntimeError('Wrong MTU value: {}'.format(mtu))
```
pavel-kirienko, line 44: See above.

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -26,25 +26,64 @@ class PythonCANMedia(_media.Media):
...
     - Usage example for Kvaser channel 0 (bitrate = 500k, mtu = 8, Node-ID = 10)::
-
-        --tr="CAN(can.media.pythoncan.PythonCANMedia('kvaser','0',5000000,8),10)"
+        --tr="CAN(can.media.pythoncan.PythonCANMedia('kvaser:0',5000000,8),10)"
     """
 
-    def __init__(self, iface_name: str, channel_name: str, bitrate: int, mtu: int) -> None:
+    VALID_FD_ADAPTER_SET = {'pcan', 'kvaser'}
+
+    def __init__(self, iface_name: str, bitrate: typing.Union[int, typing.Tuple[int, int]], mtu: int) -> None:
```
pavel-kirienko, line 38: Can you please add a few words about the arguments here?

- expected format of `iface_name`
- bitrate (tuple or int)
- mtu and its relation to bitrate (CAN FD or classic)

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -26,25 +26,64 @@ class PythonCANMedia(_media.Media):
...
         self._loopback_enabled = False
-        self._bus = can.ThreadSafeBus(interface=iface_name, channel=channel_name, bitrate=bitrate)
+        if not isinstance(bitrate, int):
+            nom_br = int(bitrate[0])
+            data_br = int(bitrate[1])
+        else:
+            nom_br = int(bitrate)
+            data_br = int(bitrate)
+        if self._is_fd == False:
+            #
```
pavel-kirienko, line 59:
```suggestion
        if not self._is_fd:
```

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -26,25 +26,64 @@ class PythonCANMedia(_media.Media):
...
+            self._bus = can.ThreadSafeBus(interface=self._conn_name[0], channel=self._conn_name[1], bitrate=nom_br)
+        elif self._conn_name[0] == 'pcan':
```
pavel-kirienko, line 61: This is not to pontificate but this design suffers from the lack of orthogonality. You have your CAN FD-capable adapters listed in `VALID_FD_ADAPTER_SET` and then you make assumptions about the contents of that set in this long `elif`. A cleaner, orthogonal design would keep pertinent elements in one place only and avoid unnecessary logical coupling. You can attain this by replacing your `VALID_FD_ADAPTER_SET` with a dict that maps the interface name (like `pcan`) to a static factory function. Then in the constructor you go like:

```python
try:
    self._bus = self.FD_ADAPTERS[self._conn_name[0]](things...)
except KeyError:
    raise can't use CAN FD on this iface
```

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -26,25 +26,64 @@ class PythonCANMedia(_media.Media):
...
+            f_clock = 40000000
+            nom_tseg1, nom_tseg2, nom_sjw = 3, 1, 1
+            data_tseg1, data_tseg2, data_sjw = 3, 1, 1
```
pavel-kirienko, line 64: Assumptions galore! Where did these numbers come from? A comment would help.

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -90,25 +129,23 @@ def configure_acceptance_filters(self, configuration: typing.Sequence[_media.Fil
...
             if f.loopback == True:
                 with self._loopback_lock:
                     self._loop_frames.append(f)
             try:
-                await self._loop.run_in_executor(self._background_executor, self.send_msg, message)
+                await self._loop.run_in_executor(self._background_executor, lambda: self._bus.send(message, timeout=monotonic_deadline - self._loop.time()))
             except asyncio.TimeoutError:
                 break
+            except:
+                _logger.exception('Can\'t send message: {}'.format(message))
```
pavel-kirienko, line 148: No.

```suggestion
```

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -142,12 +179,12 @@ def handler_wrapper(frs: typing.Sequence[_media.TimestampedDataFrame]) -> None:
...
-                                               data=frame.data,
-                                               format=frame.format,
-                                               loopback=frame.loopback,
-                                               timestamp=loop_ts))
+                            loop_ts = pyuavcan.transport.Timestamp(system_ns=time.time_ns(), monotonic_ns=time.monotonic_ns())
+                            frames.append(_media.TimestampedDataFrame(identifier=frame.identifier,
+                                          data=frame.data,
+                                          format=frame.format,
+                                          loopback=frame.loopback,
+                                          timestamp=loop_ts))
```
pavel-kirienko, line 187:
```suggestion
                            frames.append(_media.TimestampedDataFrame(identifier=frame.identifier,
                                                                      data=frame.data,
                                                                      format=frame.format,
                                                                      loopback=frame.loopback,
                                                                      timestamp=loop_ts))
```

pavel-kirienko (changes requested):
Is there any chance to add at least a basic test suite?

pavel-kirienko:
Hey @Polarisru! Please let me know if you need assistance with any of the issues I mentioned in my last review. I think it's very close to completion, just a few minor things are left to fix.

Polarisru:
Hello Pavel, sorry for this delay, I was really busy with other projects, but I will continue to process the issues on this weekend, I also would like to finish this implementation.

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
Polarisru: Hello, I am back. This magic numbers come from the settings of PCAN adapter. They don't allow any direct baudrate settings, you have to set all lengths and value of the main frequency. So I supposed these bit lengths are very universal and can be applied for almost every popular baudrate. There is probably a better solution here, but it needs significantly more time to implement it.

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
Polarisru: Was it just a formatting issue?

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
Polarisru: Yes, you ar right

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -26,25 +26,64 @@ class PythonCANMedia(_media.Media):
...
-        --tr="CAN(can.media.pythoncan.PythonCANMedia('pcan','PCAN_USBBUS1',5000000,8),10)"
-
+        --tr="CAN(can.media.pythoncan.PythonCANMedia('pcan:PCAN_USBBUS1',5000000,8),10)"
+    - Usage example for PCAN-USB channel 1 (nom.bitrate = 500k, data.bitrate = 2M, mtu = 8, Node-ID = 10)::
+        --tr="CAN(can.media.pythoncan.PythonCANMedia('pcan:PCAN_USBBUS1',[5000000,2000000],8),10)"
     - Usage example for Kvaser channel 0 (bitrate = 500k, mtu = 8, Node-ID = 10)::
-
-        --tr="CAN(can.media.pythoncan.PythonCANMedia('kvaser','0',5000000,8),10)"
+        --tr="CAN(can.media.pythoncan.PythonCANMedia('kvaser:0',5000000,8),10)"
     """
```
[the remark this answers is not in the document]
Polarisru: Must every line started with "- Usage" be followed with a blank line?

Polarisru:
I don't see any errors with mypy locally:
```
e:\Work\pyuavcan-master\pyuavcan\transport\can\media\pythoncan>mypy --ignore-missing-imports _pythoncan.py
Success: no issues found in 1 source file
```
Is it really so?

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
Polarisru: The problem is I don't know how to convert integer or tuple of integers properly without getting any error messages from mypy. If I don't check the type of baudrate with instance, mypy doesn't want to accept int().

Polarisru: And actually I don't use VALID_FD_ADAPTER_SET now

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -16,42 +24,213 @@
...
+            try:
+                frames: typing.List[_media.TimestampedDataFrame] = []
+                try:
+                    frames.append(self._read_frame())
+                except OSError as ex:
+                    raise
+                if len(self._loop_frames) > 0:
+                    with self._loopback_lock:
+                        for frame in self._loop_frames:
+                            loop_ts = pyuavcan.transport.Timestamp(system_ns=time.time_ns(), monotonic_ns=time.monotonic_ns())
```
pavel-kirienko, line 184: Please use [`pyuavcan.transport.Timestamp.now()`](https://pyuavcan.readthedocs.io/en/latest/.apidoc_generated/pyuavcan.transport.html#pyuavcan.transport.Timestamp.now) instead.

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -16,42 +24,213 @@
...
+        return _media.TimestampedDataFrame(identifier=msg.arbitration_id,
+                                           data=data,
+                                           format=frame_format,
+                                           loopback=loopback,
+                                           timestamp=timestamp)
+
+    def _set_loopback_enabled(self, enable: bool) -> None:
+        # it do nothing at the moment
+        if enable != self._loopback_enabled:
+            self._loopback_enabled = enable
```
pavel-kirienko, line 236: You don't really need this.

```suggestion
```

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -16,42 +24,213 @@
...
+            )
+        if mtu not in self.VALID_MTU_SET:
+            raise RuntimeError('Wrong MTU value: {}'.format(mtu))
+        self._mtu = int(mtu)
+        self._is_fd = self._mtu > min(self.VALID_MTU_SET) or type(bitrate) != int
+        self._loop = asyncio.get_event_loop()
+        self._closed = False
+        self._maybe_thread: typing.Optional[threading.Thread] = None
+        self._background_executor = concurrent.futures.ThreadPoolExecutor(max_workers=3)
+        self._loopback_enabled = False
```
pavel-kirienko, line 53:
```suggestion
```

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -16,42 +24,213 @@
...
+        _logger.info('Acceptance filters activated: %s', ', '.join(map(str, configuration)))
+        self._bus.set_filters(filters)
 
     async def send_until(self, frames: typing.Iterable[_media.DataFrame], monotonic_deadline: float) -> int:
-        raise NotImplementedError
+        num_sent = 0
+        for f in frames:
+            if self._closed:
+                raise pyuavcan.transport.ResourceClosedError(repr(self))
+            self._set_loopback_enabled(f.loopback)
```
pavel-kirienko, line 139:
```suggestion
```

pavel-kirienko (changes requested):
Almost there. Remaining issues:

- [x] Add a few words about the parameters in the constructor documentation, like it is done in https://github.com/UAVCAN/pyuavcan/blob/d887f1c2dea9341b7482116bcaae3230aecfd191/pyuavcan/transport/can/media/socketcan/_socketcan.py#L37-L46.

- [x] Refactor the interface construction logic -- see my suggestions.

- [ ] Get rid of the useless logic around `_loopback_enabled`.

- [ ] Some minor things I mentioned in the review.

- [ ] From my earlier review: Can we somehow make the MTU optional so that if the adapter supports CAN FD, it defaults to 64, otherwise to 8 bytes, or is it going to get convoluted? We can make the MTU argument optional and then if it's None, make it default to 8 or 64 depending on `_is_fd`.

- [ ] Add a test suite. We do not require full coverage but it is necessary to make sure that it is at least importable and somehow runnable.

pavel-kirienko:
> I don't see any errors with mypy locally:
>
>     e:\Work\pyuavcan-master\pyuavcan\transport\can\media\pythoncan>mypy --ignore-missing-imports _pythoncan.py
>     Success: no issues found in 1 source file
> Is it really so?

No. MyPy needs more context for a comprehensive analysis, so invoking it on a single file is no good. Look at the CI output:

```
pyuavcan/transport/can/media/pythoncan/_pythoncan.py: note: In member "send_until" of class "PythonCANMedia":
pyuavcan/transport/can/media/pythoncan/_pythoncan.py:147: error: Returning Any from function declared to return "object"
```

It found another error but it is not related to your code so just ignore it (that's because certain things here are not quite stable yet).

Polarisru:
* [x]  Add a few words about the parameters in the constructor documentation, like it is done in https://github.com/UAVCAN/pyuavcan/blob/d887f1c2dea9341b7482116bcaae3230aecfd191/pyuavcan/transport/can/media/socketcan/_socketcan.py#L37-L46
       .
 
     * [x]  Refactor the interface construction logic -- see my suggestions.
 
     * [x]  Get rid of the useless logic around `_loopback_enabled`.
 
     * [x]  Some minor things I mentioned in the review.
 
     * [ ]  From my earlier review: Can we somehow make the MTU optional so that if the adapter supports CAN FD, it defaults to 64, otherwise to 8 bytes, or is it going to get convoluted? We can make the MTU argument optional and then if it's None, make it default to 8 or 64 depending on `_is_fd`.
 
     * [ ]  Add a test suite. We do not require full coverage but it is necessary to make sure that it is at least importable and somehow runnable.

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -4,54 +4,343 @@
...
+
+
+def _construct_slcan(parameters: _InterfaceParameters) -> can.ThreadSafeBus:
+    if isinstance(parameters, _ClassicInterfaceParameters):
+        # only default ttyBaudrate is possible (115200)
+        return can.ThreadSafeBus(interface=parameters.interface_name,
+                                 channel=parameters.channel_name,
+                                 bitrate=parameters.bitrate)
+    elif isinstance(parameters, _FDInterfaceParameters):
+        raise TypeError('Interface does not support CAN FD: {}'.format(parameters.interface_name))
```
pavel-kirienko, line 78: This is minor but if you could update other occurrences to use f-strings that would be great. F-strings are more reliable because you can't accidentally mess up the formatting.

```suggestion
        raise TypeError(f'Interface does not support CAN FD: {parameters.interface_name}')
```

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -4,54 +4,343 @@
...
-    def __init__(self) -> None:
-        raise NotImplementedError
+    def __init__(self, iface_name: str, bitrate: typing.Union[int, typing.Tuple[int, int]], mtu: int) -> None:
+        """
+        CAN Classic/FD are possible. CAN FD is used if MTU value > 8 or two bitrates are used (nom and data).
+        
+        :param iface_name: Interface name consisting of interface and channel separated with a colon.
+            E.g., ``kvaser:0``.
+
+        :bitrate: bitrate value in bauds.
```
pavel-kirienko, line 166:
```suggestion
        :param bitrate: bitrate value in bauds. 
```

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -4,54 +4,343 @@
...
+        
+        :param mtu: The maximum data field size in bytes.
+            This value must belong to Media.VALID_MTU_SET.
+
+        """
+        #self._iface_name = str(iface_name)
+        self._conn_name = str(iface_name).split(':')
+        if len(self._conn_name) != 2:
+            raise pyuavcan.transport.InvalidMediaConfigurationError(
+                "Interface name %r does not match the format 'interface:channel'" % str(iface_name)#self._iface_name
```
pavel-kirienko, line 177: Commented out code.

```suggestion
        self._conn_name = str(iface_name).split(':')
        if len(self._conn_name) != 2:
            raise pyuavcan.transport.InvalidMediaConfigurationError(
                "Interface name %r does not match the format 'interface:channel'" % str(iface_name)
```

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -4,54 +4,343 @@
...
+        if mtu not in self.VALID_MTU_SET:
+            raise RuntimeError('Wrong MTU value: {}'.format(mtu))
+        self._mtu = int(mtu)
+        self._loop = asyncio.get_event_loop()
+        self._closed = False
+        self._maybe_thread: typing.Optional[threading.Thread] = None
+        self._background_executor = concurrent.futures.ThreadPoolExecutor(max_workers=3)
+        bitrate = (int(bitrate), int(bitrate)) if isinstance(bitrate, int) else (int(bitrate[0]), int(bitrate[1]))
+        nom_br = int(bitrate[0])
+        data_br = int(bitrate[1])
```
pavel-kirienko, line 188:
```suggestion
```

pyuavcan/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -4,54 +4,343 @@
...
+                                                       can.ThreadSafeBus]] = collections.defaultdict(
+        lambda: _construct_any, {'socketcan': _construct_socketcan,
+                                 'kvaser':    _construct_kvaser,
+                                 'slcan':     _construct_slcan,
+                                 'pcan':      _construct_pcan,
+                                 'nican':     _construct_nican,}
+    )
+
+
 class PythonCANMedia(_media.Media):
```
pavel-kirienko, line 144: Can you please move the internals towards the bottom and keep this class near the top? It's not C, no forward declarations are needed here.

pavel-kirienko (changes requested):
> Could you please give me some examples from the existing code to write a test suite?

You have two options:

1. Add `_pythoncan.py` next to https://github.com/UAVCAN/pyuavcan/blob/master/tests/transport/can/media/_socketcan.py and put the testing logic in there.

2. Keep the test close to the tested code, right in your current `pyuavcan/transport/can/media/pythoncan/_pythoncan.py` (at the bottom). In this case, you will not be able to use `@pytest.mark.asyncio` because you can't import pytest because it is not a declared dependency of the library (a user wouldn't be able to use PyUAVCAN unless PyTest is installed, which is not a reasonable constraint). Instead, you will need to create an event loop inside your test and run things manually.

In the test itself, you can open a loopback interface, write frames into it, and ensure that they arrive at the other end. Full coverage is not required. The test environment will ensure that the dependencies (PythonCAN in this case) are available so you don't need to bother about it. If necessary, you can limit the test to run only on GNU/Linux (see the linked example); by default, it will be executed on Windows also.

> Regarding the MTU: there are not so many adapters supporting CAN FD, so I am afraid of some inconvenencies using such kind of MTU definitions. We could probably use default settings and distinguish FD and Classic based on the bitrate parameter (single int or a tulpe). What do you think about it?

So you are proposing this: if the bitrate is an integer, use the default MTU of 8 bytes; if the bitrate is a tuple, default to 64 bytes? Makes sense.

pavel-kirienko:
> Could you please give me some examples from the existing code to write a test suite?

For quicker real-time communication, I recommend joining https://t.me/uavcan_ru (a public channel in Russian)

pavel-kirienko:
@Polarisru hi Alex, are you planning to continue working on this? We are only missing a basic test suite here.

Polarisru:
Yes, I have completed all other projects and want finally to fulfill this commit at this weekend. Sorry for this huge delay.

pavel-kirienko:
That's great. You will notice that the stable version of the library has a slightly different API for the CAN media layer. Please update your implementation accordingly, should be a no-brainer: https://github.com/UAVCAN/pyuavcan/commit/1b114ba00e653723f843c31b182526eea3c4bf36#diff-9ca7dfe6ff0e87b4555ef29093ef87d3c1540db8c5193ede9eb9746981cb3d5b

Also, testing is now managed via Nox instead of custom shell scripts:

    nox -s test -k pythoncan

More info at https://pyuavcan.readthedocs.io/en/latest/pages/dev.html

Polarisru:
Hello Pavel, test script is ready now. I could find a problem of malfunction but had to rebuild some features in the main python-can file, so I have to check if it still works :) I will do it tomorrow and hope we merge it finally.

pavel-kirienko:
Great. Let me help you out here a bit by rebasing your changeset to master and addressing some minor issues.

pavel-kirienko:
@Polarisru I have slightly refactored the implementation, wrote some documentation, expanded the test suite a notch, and ran a basic manual test to validate the driver (using PCAN USB and Zubax Babel). Back to you now. If you could write extra docs and add examples that would be great because what we have now is a bit minimalistic.

pavel-kirienko:
Merging, as discussed in Telegram.

pavel-kirienko (approved):

PR #126, opened 2020-10-10
Cobs

flytrex-vadim:
Switched from brute-force escaping to COBS for serial protocol serialization

### DIFF
--- a/pyuavcan/transport/serial/__init__.py
+++ b/pyuavcan/transport/serial/__init__.py
@@ -64,32 +64,29 @@ the remaining 14 least significant bits contain the service-ID value.
 Total header size: 32 bytes (256 bits).
 
 The header is prepended before the frame payload; the resulting structure is
-encoded into its serialized form using the following packet format (influenced by HDLC, SLIP, POPCOP):
+encoded into its serialized form using the following packet format:
 
 +-------------------------+--------------+---------------+--------------------------------+-------------------------+
-| Frame delimiter **0x9E**|Escaped header|Escaped payload| Escaped CRC-32C of the payload | Frame delimiter **0x9E**|
+| Frame delimiter **0x00**|Escaped header|Escaped payload| Escaped CRC-32C of the payload | Frame delimiter **0x00**|
 +=========================+==============+===============+================================+=========================+
 | 1 byte                  | 32..64 bytes | >=0 bytes     | 4..8 bytes                     | 1 byte                  |
 +-------------------------+--------------+---------------+--------------------------------+-------------------------+
-| Single-byte frame       | The following bytes are      | Four bytes long, little-endian | Same frame delimiter as |
-| delimiter **0x9E**.     | escaped: **0x9E** (frame     | byte order; bytes 0x9E (frame  | at the start.           |
-| Begins a new frame and  | delimiter); **0x8E**         | delimiter) and 0x8E (escape    | Terminates the current  |
-| possibly terminates the | (escape character). An       | character) are escaped like in | frame and possibly      |
-| previous frame.         | escaped byte is bitwise      | the header/payload. The CRC is | begins the next frame.  |
-|                         | inverted and prepended with  | computed over the unescaped    |                         |
-|                         | the escape character 0x8E.   | (i.e., original form) payload, |                         |
-|                         | For example: byte 0x9E is    | not including the header       |                         |
-|                         | transformed into 0x8E        | (because the header has a      |                         |
-|                         | followed by 0x71.            | dedicated CRC).                |                         |
+| Single-byte frame       |                              | Four bytes long, little-endian | Same frame delimiter as |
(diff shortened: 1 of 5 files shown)


pyuavcan/transport/serial/_frame.py
```diff
@@ -105,23 +105,37 @@ def compile_into(self, out_buffer: bytearray) -> memoryview:
...
 
         out_buffer[next_byte_index] = self.FRAME_DELIMITER_BYTE
         next_byte_index += 1
 
         assert (next_byte_index - 2) >= (len(header) + len(self.payload) + len(payload_crc_bytes))
         return memoryview(out_buffer)[:next_byte_index]
 
+    @staticmethod
+    def calc_cobs_size(payload_size_bytes: int) -> int:
+        return int(math.ceil(payload_size_bytes * 255.0 / 254.0))
```
pavel-kirienko, line 125: It is better to avoid floating point when dealing with discrete entities like bytes for simplicity and to provide cleaner implementation examples for people who will be re-implementing this logic for embedded systems.

```suggestion
        return (payload_size_bytes * 255 + 253) // 254
```

This change is equivalent which can be verified:

```python
>>> set(int(math.ceil(payload_size_bytes * 255.0 / 254.0)) - ((payload_size_bytes * 255 + 253) // 254) for payload_size_bytes in range(10000000))
{0}
```

The `math` import is no longer needed.

pyuavcan/transport/serial/__init__.py
```diff
@@ -64,32 +64,29 @@
...
 
 The header is prepended before the frame payload; the resulting structure is
-encoded into its serialized form using the following packet format (influenced by HDLC, SLIP, POPCOP):
+encoded into its serialized form using the following packet format:
 
 +-------------------------+--------------+---------------+--------------------------------+-------------------------+
-| Frame delimiter **0x9E**|Escaped header|Escaped payload| Escaped CRC-32C of the payload | Frame delimiter **0x9E**|
+| Frame delimiter **0x00**|Escaped header|Escaped payload| Escaped CRC-32C of the payload | Frame delimiter **0x00**|
 +=========================+==============+===============+================================+=========================+
 | 1 byte                  | 32..64 bytes | >=0 bytes     | 4..8 bytes                     | 1 byte                  |
```
pavel-kirienko, line 72: With COBS the header and the CRC have fixed size now.

```suggestion
| 1 byte                  | 32 bytes     | >=0 bytes     | 4 bytes                        | 1 byte                  |
```

pavel-kirienko (changes requested):
[From the forum](https://forum.uavcan.org/t/uavcan-serial-issues-with-dma-friendliness-and-bandwidth-overhead/846/48?u=pavel.kirienko):

> At the whole pyuavcan level there are some dsdl test failures that are probably setup-related (PYTHONASYNCIODEBUG)

You should be able to fix that by exporting `PYTHONASYNCIODEBUG=1`, or just by running `test.sh`.

Some of the docs make references to escaping which may need reviewing. For example, the docs for the method `SerialFrame.compile_into` talk about escaping, is it still valid?

The docs would benefit if you added a link to https://forum.uavcan.org/t/uavcan-serial-issues-with-dma-friendliness-and-bandwidth-overhead/846 because that thread provides an exhaustive explanation of why COBS has been chosen.

pyuavcan/transport/serial/_stream_parser.py
```diff
@@ -64,13 +64,6 @@ def _process_byte(self, b: int, timestamp: pyuavcan.transport.Timestamp) -> None
             return
 
         # Unescaping is done only if we're inside a frame currently.
```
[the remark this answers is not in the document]
flytrex-vadim: the docs for the method SerialFrame.compile_into reads valid to me, considering COBS is also an escaping of sorts.

flytrex-vadim:
The docs for the method SerialFrame.compile_into reads valid to me, considering COBS is also an escaping of sorts.
Other remarks fixed.

pyuavcan/transport/serial/_frame.py
```diff
@@ -7,9 +7,9 @@
 from __future__ import annotations
 import typing
 import struct
-import itertools
 import dataclasses
 import pyuavcan
+from cobs import cobs
```
pavel-kirienko, line 12: This is to suppress the MyPy warning about the type hints being unavailable.

```suggestion
from cobs import cobs    # type: ignore
```

pavel-kirienko (changes requested):
There is one issue found by MyPy. Please accept my suggestion, then we should be good to go. The other issues from MyPy are caused by the fact that its version number is not yet frozen (because the library is still being actively developed); a new one was released which found extra issues in unrelated code. I should freeze it soon.

pyuavcan/transport/serial/_frame.py
flytrex-vadim: yes, just committed this fix

pavel-kirienko (approved):

PR #139, opened 2020-12-24
Remove CLI

### LINKED ISSUE #134: Extract the CLI tool into a separate package/repository
Contrary to my expectations, the CLI tool has little in common with the library so keeping it in the same codebase is not sensible. The addition of new higher-level functionality in the future would only make it harder to justify having a shared package for both.

pavel-kirienko:
Fixes #134 

The new repository for the CLI is being worked on, stay tuned. This PR should be merged once it is ready.

The integration test for the demo application had to be rewritten because the old one was built on top of the CLI.

The overall directory layout has been altered: the demo script and the DSDL namespaces it depends on are now located directly under the project root at `/demo/`. This should make the demo more accessible.

The coverage scaffolding was changed: now we use `coverage run ...` directly instead of relying on `sitecustomize.py`. This is simpler and easier to maintain.

The integration test suite does no longer generate DSDL packages in the project root (I've been looking for an easy way to fix it for a while).

This is the last substantial changeset before the v1.1 release.


pavel-kirienko:
Done, the new tool is available for use. Read the announcement here: https://forum.uavcan.org/t/yakut-the-new-name-for-pyuavcan-cli-tool/1049

Issue #166, opened 2021-05-10
[UAVCAN v0] Release the legacy library as `pyuavcan_v0` to avoid compatibility issues


pavel-kirienko:
See "Legacy considerations" at https://pyuavcan.readthedocs.io/en/stable/pages/installation.html#legacy-considerations

Context at https://forum.uavcan.org/t/problems-with-ds-015/1219/41?u=pavel.kirienko

FYI @tridge

PrincePepper:
Tell me, is there a big difference between `uavcan` and `pyuavcan_v0` now, or has only the name changed?

pavel-kirienko:
Only the name has changed.

PrincePepper:
Thanks for the answer! In the meantime, I'm too lazy to change in the project 😂

pavel-kirienko:
Unless you require compatibility with ArduPilot, you should really switch to UAVCAN v1, which is fully supported by [PyUAVCAN v1.2](https://pyuavcan.readthedocs.io).

PR #174, opened 2021-07-24
Remove section "legacy considerations" from the installation docs

pavel-kirienko:
Because #166 is fixed


docs/pages/installation.rst
```diff
@@ -40,18 +40,3 @@ and make sure to include at least its core dependencies, which are:
...
-which implements an early experimental version of the protocol known as UAVCAN v0
-that is no longer recommended for new designs.
-It should not be confused with this library which implements the long-term stable version
-of the protocol known as UAVCAN v1.
-
-Having both ``pyuavcan`` and the old ``uavcan`` libraries installed in the same environment is not recommended,
-because PyUAVCAN generates Python packages from DSDL namespaces,
-and since the standard DSDL types are stored in the root namespace named ``uavcan``,
-it would conflict with the old library.
-Hence, when installing PyUAVCAN, make sure you don't have the legacy library around: ``pip uninstall -y uavcan``.
```
coderkalyan, line 57: It is possible that someone still has the `uavcan` package from v0 installed. I suggest that you replace this with a short note saying that that package will indeed conflict, but users can install `pyuavcan_v0` to mitigate this.

coderkalyan (changes requested):

docs/pages/installation.rst
pavel-kirienko: I pushed a new (and last, presumably) update to the legacy `uavcan` package that removes the actual module from the system when updated, so it doesn't conflict anymore. In the worst case, we will provide guidance on the forum if there are still users affected by this.

coderkalyan: Alright.

coderkalyan (approved):

PR #185, opened 2021-09-18
socketcan: improve incorrect config error message

### LINKED ISSUE #183: UAVCAN/CAN: SocketCAN: implement better error reporting when sending a CAN FD frame over a Classic CAN socket
Details: https://forum.uavcan.org/t/yakut-error-oserror-errno-22-invalid-argument/1434

Fix this:

https://github.com/UAVCAN/pyuavcan/blob/9da3cf579b65cae4067b16c621cc3321156f17d4/pyuavcan/transport/can/media/socketcan/_socketcan.py#L127-L133

coderkalyan:
Closes #183 

Currently, as per Python exception chaining, the OSError is still thrown, but display of the MediaConfigurationError takes precedence.

### DIFF
--- a/pyuavcan/transport/can/media/socketcan/_socketcan.py
+++ b/pyuavcan/transport/can/media/socketcan/_socketcan.py
@@ -129,6 +129,14 @@ class SocketCANMedia(Media):
                     self._loop.sock_sendall(self._sock, self._compile_native_frame(f.frame)),
                     timeout=monotonic_deadline - self._loop.time(),
                 )
+            except OSError as err:
+                if err.errno == errno.EINVAL and self._is_fd:
+                    raise pyuavcan.transport.InvalidMediaConfigurationError(
+                        "Invalid socketcan configuration: "
+                        "the device probably doesn't support CAN-FD. "
+                        "Try setting MTU to 8 (Classic CAN)"
+                    ) from err
+                raise err
             except asyncio.TimeoutError:
                 break
             else:


pyuavcan/transport/can/media/socketcan/_socketcan.py
```diff
@@ -129,6 +129,8 @@ async def send(self, frames: typing.Iterable[Envelope], monotonic_deadline: floa
                     self._loop.sock_sendall(self._sock, self._compile_native_frame(f.frame)),
                     timeout=monotonic_deadline - self._loop.time(),
                 )
+            except OSError as err:
+                raise pyuavcan.transport.InvalidMediaConfigurationError("Invalid socketcan configuration") from err
```
pavel-kirienko, line 133: OSError can mean a lot of things. I would add an explicit check here like (requires `import errno`):

    if err.errno == errno.EINVAL and self._is_fd:
        ...

...and raise the MediaConfigurationError only if the check holds (otherwise just re-`raise` it as-is). Also, maybe the text should go like `Invalid socketcan configuration: the device probably does not support CAN FD; try Classic CAN` or something similar, the point is to add more information to aid the user in solving this problem.

pavel-kirienko (changes requested):
Can we maybe move your work on the current issues into a separate `development` branch and then merge them all at once? This is important for PyUAVCAN because we have continuous deployment here, so that every little change like this one would necessitate a separate patch version number.

coderkalyan:
@pavel-kirienko changed to point to the develop branch, and resolved the comment you had.

pyuavcan/transport/can/media/socketcan/_socketcan.py
```diff
@@ -129,6 +129,11 @@ async def send(self, frames: typing.Iterable[Envelope], monotonic_deadline: floa
                     self._loop.sock_sendall(self._sock, self._compile_native_frame(f.frame)),
                     timeout=monotonic_deadline - self._loop.time(),
                 )
+            except OSError as err:
+                if err.errno == errno.EINVAL and self._is_fd:
+                    raise pyuavcan.transport.InvalidMediaConfigurationError("Invalid socketcan configuration: the device probably doesn't support CAN-FD. Try setting MTU to 8 (Classic CAN)") from err
```
pavel-kirienko, line 134: This line is too long and won't pass the style check.

pyuavcan/transport/can/media/socketcan/_socketcan.py
```diff
@@ -129,6 +129,11 @@ async def send(self, frames: typing.Iterable[Envelope], monotonic_deadline: floa
...
+                else:
+                    raise err
```
pavel-kirienko, line 136: Redundant `else` (pylint will complain about this).

```suggestion
                raise err
```

pavel-kirienko (changes requested):

pyuavcan/transport/can/media/socketcan/_socketcan.py
coderkalyan: Apologies, didn't notice. Fixed.

pavel-kirienko (approved):

PR #187, opened 2021-09-29
Mtu8

### LINKED ISSUE #184: [UAVCAN/CAN]: PythonCAN media driver should choose Classic CAN if MTU=8 and bitrates are equal
This heuristic should be implemented here:

https://github.com/UAVCAN/pyuavcan/blob/9da3cf579b65cae4067b16c621cc3321156f17d4/pyuavcan/transport/can/media/pythoncan/_pythoncan.py#L155

Currently, it is not possible to use `pyuavcan.application.make_transport()` with SLCAN because it attempts to initialize it in FD mode which expectedly fails with an error "interface does not support CAN FD".

coderkalyan:
Closes #184


pavel-kirienko:
I recommend configuring a File Watcher to automatically run Black on save -- it helps avoid broken builds like you have here caused by misformatted sources.

coderkalyan:
Yes, you're right, I should. I should also be more careful and review diffs before pushing. Apologies.

coderkalyan:
Closes #184

pavel-kirienko:
PSA: I see that you use force-pushing, presumably to maintain clean history. This is admirable but not really necessary in our repositories because we use squash merging -- all your commits will be squashed into one when merging anyway. This is just FYI.

coderkalyan:
> PSA: I see that you use force-pushing, presumably to maintain clean history. This is admirable but not really necessary in our repositories because we use squash merging -- all your commits will be squashed into one when merging anyway. This is just FYI.

Thanks for the heads up. Yes, this is a habit I've grown from other repositories that use rebase merging rather than squash. I'll keep it in mind.

Issue #176, opened 2021-08-05
UAVCAN/serial: Reception of malformed frames may cause the transport instance to raise unexpected errors and self-terminate


pavel-kirienko:
Example:

```
2021-08-05 18:45:24 0115854 ERR pyuavcan.transport.serial._serial: SerialTransport('socket://localhost:50905', local_node_id=None, service_transfer_multiplier=2, baudrate=9600): Reader thread has failed, the instance with port Serial<id=0x7fcf0bc5b970, open=True>(port='socket://localhost:50905', baudrate=9600, bytesize=8, parity='N', stopbits=1, timeout=1.0, xonxoff=False, rtscts=False, dsrdtr=False) will be terminated: Invalid subject ID: 32741                      
Traceback (most recent call last):
  File "/home/pavel/.local/lib/python3.9/site-packages/pyuavcan/transport/serial/_serial.py", line 439, in _reader_thread_func
    parser.process_next_chunk(chunk, chunk_ts)
  File "/home/pavel/.local/lib/python3.9/site-packages/pyuavcan/transport/serial/_stream_parser.py", line 55, in process_next_chunk
    self._finalize(known_invalid=self._outside_frame)
  File "/home/pavel/.local/lib/python3.9/site-packages/pyuavcan/transport/serial/_stream_parser.py", line 80, in _finalize
    parsed = SerialFrame.parse_from_cobs_image(buf)
  File "/home/pavel/.local/lib/python3.9/site-packages/pyuavcan/transport/serial/_frame.py", line 134, in parse_from_cobs_image
    return SerialFrame.parse_from_unescaped_image(memoryview(unescaped_image))
  File "/home/pavel/.local/lib/python3.9/site-packages/pyuavcan/transport/serial/_frame.py", line 171, in parse_from_unescaped_image
    data_specifier = pyuavcan.transport.MessageDataSpecifier(int_data_spec)
  File "<string>", line 4, in __init__
  File "/home/pavel/.local/lib/python3.9/site-packages/pyuavcan/transport/_data_specifier.py", line 26, in __post_init__
    raise ValueError(f"Invalid subject ID: {self.subject_id}")
ValueError: Invalid subject ID: 32741
```

The solution is to catch ValueError raised from the frame parser method and return None (this behavior is specified in the API contract but implemented incorrectly):

[...]

PR #189, opened 2021-11-03
transport: serial: catch malformed frames

coderkalyan:
Pushed this real quick after reading #176 but wasn't sure how to test it @pavel-kirienko? What's the easiest way to induce the correct type of serial error?

### DIFF
--- a/pyuavcan/transport/serial/_frame.py
+++ b/pyuavcan/transport/serial/_frame.py
@@ -131,7 +131,10 @@ class SerialFrame(pyuavcan.transport.commons.high_overhead_transport.Frame):
             unescaped_image = cobs.decode(bytearray(image))  # TODO: PERFORMANCE WARNING: AVOID THE COPY
         except cobs.DecodeError:
             return None
-        return SerialFrame.parse_from_unescaped_image(memoryview(unescaped_image))
+        try:
+            SerialFrame.parse_from_unescaped_image(memoryview(unescaped_image))
+        except ValueError:
+            return None
 
     @staticmethod
     def parse_from_unescaped_image(header_payload_crc_image: memoryview) -> typing.Optional[SerialFrame]:


pyuavcan/transport/serial/_frame.py
```diff
@@ -131,7 +131,10 @@ def parse_from_cobs_image(image: memoryview) -> typing.Optional[SerialFrame]:
             unescaped_image = cobs.decode(bytearray(image))  # TODO: PERFORMANCE WARNING: AVOID THE COPY
         except cobs.DecodeError:
             return None
-        return SerialFrame.parse_from_unescaped_image(memoryview(unescaped_image))
+        try:
+            SerialFrame.parse_from_unescaped_image(memoryview(unescaped_image))
```
pavel-kirienko, line 135: `return` statement is missing, but the more important thing is that this exception block seems to be a little out of place: the docs for `parse_from_unescaped_image()` state that it should return None for invalid inputs but the actual behavior is that it throws. Should we not move the exception block there?

pavel-kirienko (changes requested):
>What's the easiest way to induce the correct type of serial error?

I recommend adding a new test case based on this one:

https://github.com/UAVCAN/pyuavcan/blob/9da3cf579b65cae4067b16c621cc3321156f17d4/pyuavcan/transport/serial/_frame.py#L319-L320

Except that you make an invalid header and then check that None is returned (consult with [`__post_init__()`](https://github.com/UAVCAN/pyuavcan/blob/76f29d67938a8f3dc6a44eb1c812049fdd757f2e/pyuavcan/transport/serial/_frame.py#L46-L63) to see which headers are considered invalid).

pavel-kirienko:
I am going to take this over since I am currently working on a minor revision fixing a few things.

PR #196, opened 2021-12-17
Legacy v0.5 with CAN-FD support in SocketCAN

emrainey:
Branching Legacy PyUAVCAN to v0.5 and adding support to CAN-FD 2.0B frames in SocketCAN as well as support for TAO. 

A new legacy-v0.5 branch should be maintained separate from legacy-v0.


emrainey:
@thirtytwobits , can you take a look at this?

emrainey:
Since there's not a base branch `legacy-v0.5` in the repo now, this will attempt to merge to `legacy-v0` which is not strictly what we would want I guess.

thirtytwobits:
we'll need a new legacy-v0.5 branch upstream to be able to accept this.
@pavel-kirienko , is it okay if I create the .05 branch? (Erik is with me :)

pavel-kirienko:
Okay. What is the expected lifespan of this new legacy branch? Who is the owner of the PyPI release? I noticed a new repository secret introduced with this changeset but I don't suppose it is defined yet, is it?

thirtytwobits:
the lifespan is up to you. We're just contributing back more of the CAN-FD changes we use internally. We don't have to do a PyPI release unless you wanted to. It's just good to have this stay consistent with the v0 branch

pavel-kirienko:
Okay. I guess we could just accept it as-is (on a new branch) in case this is ever needed later, but I don't really expect it to be used at all. This is why I asked about the lifespan. This is such a deep legacy so troubled with technical debt, I don't dare to look at it if I can help it.

thirtytwobits:
I have created a legacy-v0.5 branch. Please re-submit this PR as targeting that branch.

emrainey:
Resubmitting

PR #210, opened 2022-02-22
Add cantact interface

chemicstry:
This adds `cantact` interface from `python-can`

### DIFF
--- a/pyuavcan/transport/can/media/pythoncan/_pythoncan.py
+++ b/pyuavcan/transport/can/media/pythoncan/_pythoncan.py
@@ -451,6 +451,16 @@ def _construct_canalystii(parameters: _InterfaceParameters) -> can.ThreadSafeBus
     assert False, "Internal error"
 
 
+def _construct_cantact(parameters: _InterfaceParameters) -> can.ThreadSafeBus:
+    if isinstance(parameters, _ClassicInterfaceParameters):
+        return can.ThreadSafeBus(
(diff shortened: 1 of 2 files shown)


chemicstry:
Note that `python-can` v4.0 is now released

pavel-kirienko (changes requested):
Thanks. Please [fix the formatting as indicated by CI](https://ci.appveyor.com/project/Zubax/pyuavcan/builds/42674727/job/yvr6082hpwqo86ji#L645) and bump the minor version number.

pavel-kirienko (approved):

chemicstry:
I think the CI got stuck, but could you cancel the auto merge? I think the issue https://github.com/UAVCAN/yakut/issues/43 is related to the cantact driver in `python-can`, but I'm still investigating.

pavel-kirienko:
Okay. Let me know how it goes.

chemicstry:
So I've confirmed that there is a problem with `cantact` driver in `python-can`. For some reason it causes the python runtime to completely slow down, even exception traces are slow to print. I believe it messes things up somewhere in the FFI layer.

The alternative `gs_usb` driver in `python-can` also works with cantact adapters and so far it seems to work fine with pyuavcan/yakut. However, it has a very inconvenient API and requires USB bus/address to initialize, which changes on each USB enumeration. So in order to be "usable", pyuavcan would need usb dependency to do a lookup based on vid/pid, or I could submit a PR to python-can to add the scan function there instead.

I will close this PR for now.

PR #212, opened 2022-03-03
Implement gs_usb interface

chemicstry:
Another attempt on supporting `Canable` adapters by implementing `gs_usb` interface after #210 failed.

This is currently blocked on upstream changes: https://github.com/hardbyte/python-can/pull/1270

I was not sure how to go about implementing hardware loopback frames. Specifying boolean for each interface seemed too verbose. Instead, I made it switch to hardware loopback as soon as it receives a TX frame.

This seems to be working fine with yakut monitor/call/subscribe on windows and canable adapter with candleLight firmware.


pavel-kirienko:
Okay. Please tag me when unblocked.

chemicstry:
One test fails [here](https://github.com/UAVCAN/pyuavcan/blob/510feed778df657d64f4054912f15f490df9acb4/tests/transport/can/media/_pythoncan.py#L143), but it's because `python-can` uses the `MSG_DONTROUTE` flag to detect loopback, unlike `MSG_CONFIRM` in `pyuavcan`'s socketcan implementation. This makes all virtual can frames to be loopback (`is_rx` false), because they originate from the same machine. Fixing this in `python-can` could break someone's else code, depending on the use case. So IMO, the best way is to either disable the loopback assert or implement an exception for virtual socket can.

pavel-kirienko:
>So IMO, the best way is to either disable the loopback assert or implement an exception for virtual socket can.

This is not just an assertion check, this is a unit test. Removing the test will make it pass but break the production code.

pavel-kirienko:
@chemicstry any update here? do we still want this open?

chemicstry:
There was a big discussion in `python-can` repository and TLDR of it is that rx/echo/loopback/confirm/etc terminology is completely mixed up and I don't think I personally can resolve it in the `python-can` library. Properly fixing it would be a breaking change and I'm not sure if anyone is up to it.

An alternative without touching `python-can` is to conditionally enable hardware loopback just for select interfaces and I think that is the way to go. Unfortunatelly, my PR for the required `gs_usb` interface fixes is still hanging in `python-can`, I'll try pinging.

pycyphal/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -22,6 +22,16 @@
 _logger = logging.getLogger(__name__)
 
 
+class PythonCANBusOptions:
```
pavel-kirienko, line 25: Please make this a dataclass.

pavel-kirienko (changes requested):
Can you please add your remark regarding the possibility of enabling the hardware loopback for the other interfaces to the class docstring?

chemicstry:
Updated docs and while at it also realized that gs_usb support hardware monotonic timestamps. Added that as well.

pycyphal/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -313,8 +337,17 @@ def _read_batch(self) -> typing.List[typing.Tuple[Timestamp, Envelope]]:
             msg = self._bus.recv(0.0 if batch else self._MAXIMAL_TIMEOUT_SEC)
             if msg is None:
                 break
-            timestamp = Timestamp.now()  # TODO: use accurate timestamping
-            loopback = False  # TODO: no possibility to get real loopback yet
+
+            if self._bus_options.hardware_timestamp:
+                timestamp = Timestamp(system_ns=time.time_ns(), monotonic_ns=Timestamp._second_to_ns(msg.timestamp))
```
pavel-kirienko, line 342: Uh-oh. Should we make `_second_to_ns` public or should we manually multiply the timestamp by 1e9 here? Reaching for a private method like this gives me pain.

pycyphal/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -313,8 +337,17 @@ def _read_batch(self) -> typing.List[typing.Tuple[Timestamp, Envelope]]:
...
+            else:
+                timestamp = Timestamp.now()
+
+            if self._bus_options.hardware_loopback:
+                loopback = not msg.is_rx
+            else:
+                loopback = False
```
pavel-kirienko, line 349: You like conditional branching I see. Okay.

pycyphal/transport/can/media/pythoncan/_pythoncan.py
chemicstry: I'm not too familiar with python and thought it is a package-private method. Neither python nor linter complained though :D I think multiplying by 1e9 should be clear enough

pavel-kirienko: I am also wondering why the linter didn't complain. I am used to PyCharm yelling at me when I do something like this so I didn't even check if PyLint is configured to report issues of this kind.

pavel-kirienko (approved):

chemicstry:
> > But this still has to wait for the new python-can release
> 
> @chemicstry do we have any eta?

I don't know, probably not very soon, `python-can` is really short on maintainers.

Although if this was merged it wouldn't cause any issues unless `gs_usb` interface was selected, then you would get interface not found error. But you could still use it if you installed `python-can` from git.

Since this contains more relevant changes for other interfaces (`PythonCANMediaOptions`), maybe adding a note to `gs_usb` docs that it requires latest `python-can` from git would be fine for now?

PR #231, opened 2022-06-23
Fix gs_usb initialization

chemicstry:
Sorry for a dumb mistake, introduced with the last commit adding python-can version warnings. I swear it was the formatter that added these parentheses 🙄


pavel-kirienko (changes requested):
Please set the version number to 1.9.0b1 and I'll merge it right away.

pavel-kirienko (approved):

Issue #233, opened 2022-07-06
OSError No buffer space available during firmware update / large data transfer with socketcan


wiboticalex:
Occasionally during a firmware update Yakut will crash and the socket that pycyphal has created will claim that there is "No buffer space available" aka ENOBUFS.

Traceback:
```
2022-07-05 18:02:02 0101068 ERR pycyphal.presentation._port._server: Server(dtype=uavcan.file.Read.1.1, input_transport_session=CANInputSession(InputSessionSpecifier(data_specifier=ServiceDataSpecifier(service_id=408, role=<Role.REQUEST: 1>), remote_node_id=None), PayloadMetadata(extent_bytes=300))) task failure: [Errno 105] No buffer space available
Traceback (most recent call last):
  File "/home/alex/.local/share/virtualenvs/uavcan_v1-Kt-sRht1/lib/python3.9/site-packages/pycyphal/presentation/_port/_server.py", line 201, in task_function
    await self.serve_for(handler, _LISTEN_FOREVER_TIMEOUT)
  File "/home/alex/.local/share/virtualenvs/uavcan_v1-Kt-sRht1/lib/python3.9/site-packages/pycyphal/presentation/_port/_server.py", line 182, in serve_for
    return await self.serve(handler, monotonic_deadline=loop.time() + timeout)
  File "/home/alex/.local/share/virtualenvs/uavcan_v1-Kt-sRht1/lib/python3.9/site-packages/pycyphal/presentation/_port/_server.py", line 170, in serve
    await self._do_send(response, meta, response_transport_session, loop.time() + self._send_timeout)
  File "/home/alex/.local/share/virtualenvs/uavcan_v1-Kt-sRht1/lib/python3.9/site-packages/pycyphal/presentation/_port/_server.py", line 309, in _do_send
    return await session.send(transfer, monotonic_deadline)
  File "/home/alex/.local/share/virtualenvs/uavcan_v1-Kt-sRht1/lib/python3.9/site-packages/pycyphal/transport/can/_session/_output.py", line 252, in send
    return await self._do_send(can_id, transfer, monotonic_deadline)
  File "/home/alex/.local/share/virtualenvs/uavcan_v1-Kt-sRht1/lib/python3.9/site-packages/pycyphal/transport/can/_session/_output.py", line 163, in _do_send
    if await self._send_handler(transaction):
[...]

PR #234, opened 2022-07-07
Avoid ENOBUFS OSErrors by shrinking the size of SNDBUF when appropriate.

wiboticalex:
This is a followup to issue #233.

Avoid ENOBUFS OSErrors by shrinking the size of SNDBUF when appropriate.

Similar behavior could have probably been achieved by simply setting the socket's SNDBUF to 0 to get whatever Linux sets as the minimum, but it seems like this may cause potential queue under-utilization when there are large queues specified, so this pull request tries to calculate (with a very rough approximation for sk_buffer struct size) what the minimum SNDBUF should be to get blocking behavior.

### DIFF
--- a/pycyphal/transport/can/media/socketcan/_socketcan.py
+++ b/pycyphal/transport/can/media/socketcan/_socketcan.py
@@ -14,6 +14,7 @@ import logging
 import warnings
 import threading
 import contextlib
+import pathlib
 import pycyphal.transport
 from pycyphal.transport import Timestamp
 from pycyphal.transport.can.media import Media, Envelope, FilterConfiguration, FrameFormat
@@ -79,7 +80,7 @@ class SocketCANMedia(Media):
         )
         self._native_frame_size = _FRAME_HEADER_STRUCT.size + self._native_frame_data_capacity
 
-        self._sock = _make_socket(iface_name, can_fd=self._is_fd)
+        self._sock = _make_socket(iface_name, can_fd=self._is_fd, native_frame_size=self._native_frame_size)
         self._ctl_main, self._ctl_worker = socket.socketpair()  # This is used for controlling the worker thread.
         self._closed = False
         self._maybe_thread: typing.Optional[threading.Thread] = None
@@ -309,6 +310,7 @@ _TIMEVAL_STRUCT = struct.Struct("@Ll")  # Using native size because the native d
 
 # From the Linux kernel; not exposed via the Python's socket module
 _SO_TIMESTAMP = 29
+_SO_SNDBUF = 7
 
 _CANFD_BRS = 1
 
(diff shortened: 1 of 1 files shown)


pycyphal/transport/can/media/socketcan/_socketcan.py
```diff
@@ -319,11 +321,32 @@ class _NativeFrameDataCapacity(enum.IntEnum):
...
 
 
-def _make_socket(iface_name: str, can_fd: bool) -> socket.socket:
+def _get_tx_queue_len(iface_name: str) -> int:
+    try:
+        sysfs_net = pathlib.Path("/sys/class/net/")
+        sysfs_tx_queue_len = sysfs_net / iface_name / "tx_queue_len"
+        return int(sysfs_tx_queue_len.read_text())
+    except FileNotFoundError as e:
+        raise FileNotFoundError("tx_queue_len sysfs location not found") from e
```
pavel-kirienko, line 330: What is the point of catching and raising the same exception here? Are you sure we need this?

pycyphal/transport/can/media/socketcan/_socketcan.py
```diff
@@ -319,11 +321,32 @@ class _NativeFrameDataCapacity(enum.IntEnum):
...
     s = socket.socket(socket.PF_CAN, socket.SOCK_RAW, socket.CAN_RAW)  # type: ignore
     try:
         s.bind((iface_name,))
         s.setsockopt(socket.SOL_SOCKET, _SO_TIMESTAMP, 1)  # timestamping
+        default_sndbuf_size = s.getsockopt(socket.SOL_SOCKET, _SO_SNDBUF)
+
+        # approximate sk_buffer kernel struct overhead.
+        # A lower estimate over higher estimate is preferred since _SO_SNDBUF will enforce
+        # a minimum value, and blocking behavior will not work if this is too high.
+        SKB_OVERHEAD = 444
```
pavel-kirienko, line 343: Maybe it's better to move this to the module scope as `_SKB_OVERHEAD`?

pycyphal/transport/can/media/socketcan/_socketcan.py
```diff
@@ -319,11 +321,32 @@ class _NativeFrameDataCapacity(enum.IntEnum):
...
+
+        blocking_sndbuf_size = (native_frame_size + SKB_OVERHEAD) * _get_tx_queue_len(iface_name)
+
+        # Allow CAN sockets to block when full similar to how Ethernet sockets do.
+        # Avoids ENOBUFS errors on TX when queues are full in most cases.
```
pavel-kirienko, line 348:
```suggestion
        # Avoids ENOBUFS errors on TX when queues are full in most cases.
        # More info:
        #   - https://github.com/OpenCyphal/pycyphal/issues/233
        #   - "SocketCAN and queueing disciplines: Final Report", Sojka et al, 2012
```

pavel-kirienko (changes requested):
Please also bump the version number in `pycyphal/VERSION` to `1.9.0.beta2` and (optionally) add a line to `CHANGELOG.rst` about this.

pavel-kirienko (approved):

pycyphal/transport/can/media/socketcan/_socketcan.py
pavel-kirienko: Okay, I see.

pavel-kirienko:
@wiboticalex can you please run Black to fix the formatting? This should get merged automatically afterward.

wiboticalex:
Done. Thought I could get away with moving code around and not reformatting, but apparently not :slightly_smiling_face: .

Thanks @pavel-kirienko

Issue #227, opened 2022-04-30
Properly handle global state changes like network interface becoming unavailable


pavel-kirienko:
This is perhaps related to https://github.com/OpenCyphal/pycyphal/issues/224

When the network interface becomes unavailable (e.g., CAN adapter disconnected, IP address changed, Cyphal/serial TCP connection closed), all services dependent on the active transport instance should be notified such that the state change can be adequately processed.

PR #235, opened 2022-07-15
Add an experimental Cyphal/CAN media implementation for replaying candump files, release v1.9

pavel-kirienko:
Also, update linters to the latest versions.

Usage example with Yakut:

```bash
export UAVCAN__CAN__IFACE='candump:verification/integration/candump.log'
y sub uavcan.node.heartbeat 10:reg.udral.service.common.readiness 130:reg.udral.service.actuator.common.status
y mon
```

Currently, by default, the upper layers will not be notified when the end of the log file is reached. This may be worked around later as part of #227. Meanwhile, you can instruct CandumpMedia to terminate its process when the end of the file is reached by exporting `PYCYPHAL_CANDUMP_YOU_ARE_TERMINATED=1`.

Test against Python 3.9 fails randomly due to problems with AppVeyor, so disregard that.

cc @silverv


pavel-kirienko:
https://github.com/appveyor/ci/issues/3832

PR #236, opened 2022-08-11
Initial automatic dsdl compilation

### LINKED ISSUE #153: Implement implicit DSDL compilation via import hooks
Instead of this:

```python
import pyuavcan
compiled_dsdl_dir = pathlib.Path(__file__).resolve().parent / ".demo_dsdl_compiled"
sys.path.insert(0, str(compiled_dsdl_dir))
try:
    import sirius_cyber_corp
    import pyuavcan.application
except (ImportError, AttributeError):
    src_dir = pathlib.Path(__file__).resolve().parent
    pyuavcan.dsdl.compile_all(
        [
            src_dir / "custom_data_types/sirius_cyber_corp",
            src_dir / "public_regulated_data_types/uavcan/",
        ],
        output_directory=compiled_dsdl_dir,
    )
    importlib.invalidate_caches()
    import sirius_cyber_corp
    import pyuavcan.application
```

We could just do this:

```python
import pyuavcan
import sirius_cyber_corp
import pyuavcan.application
```

Such that if the DSDL namespaces are not compiled, PyUAVCAN would invoke the compiler transparently for the user. This is implementable with the help of [Python import hooks](https://docs.python.org/3/reference/import.html#import-hooks). The Cap'n'Proto Python client implements it this way.

We could define an environment variable named simply `CYPHALPATH` listing a set of paths where DSDL namespaces are to be searched for. When an import hook is triggered, PyUAVCAN would scan the paths looking for the matching namespace. If found, it would invoke the DSDL compiler, providing all other directories reachable via the path variable as the look-up paths. The output should be cached somewhere secure to avoid recompilation at every launch (`/tmp` is probably not an option for security reasons, but the home directory might work).

chemicstry:
This implements automatic DSDL compilation and fixes #153.

Opening this as an incomplete draft PR so see if I'm going in a correct direction.

Todo:
- [x] Load default paths from environment variables. 
- [ ] Check for changed source files. Can be revisited later.
- [x] Documentation.
- [x] Remove temporary debug scripts and lines.


pavel-kirienko:
This is nice. As far as I understand, the key missing bit is removing the need to install the import hook manually --- it should be done automatically when PyCyphal is imported. The values for `lookup_directories` are to be sourced from `CYPHALPATH`, and `output_directory` is similarly initialized from `PYCYPHALPATH` (with a default), right? We may want to initialize `allow_unregulated_fixed_port_id` similarly to `CYPHAL_ALLOW_UNREGULATED_FIXED_PORT_ID` but it is far less important.

If I remember our conversation correctly, there will be no default value for `CYPHALPATH` (i.e., empty), but for `PYCYPHALPATH` the default would be either `~/.pycyphal` (or some equivalent on non-GNU platforms) or `$CYPHALPATH[0]/.pycyphal`. Which one should we go with?

chemicstry:
> This is nice. As far as I understand, the key missing bit is removing the need to install the import hook manually --- it should be done automatically when PyCyphal is imported. The values for `lookup_directories` are to be sourced from `CYPHALPATH`, and `output_directory` is similarly initialized from `PYCYPHALPATH` (with a default), right? We may want to initialize `allow_unregulated_fixed_port_id` similarly from `CYPHAL_ALLOW_UNREGULATED_FIXED_PORT_ID` but it is far less important.

Yes, although might I suggest to use `CYPHAL_PATH` and `PYCYPHAL_PATH`? Seems a bit more standard, at least from my experience. Unless `CYPHALPATH` is already in use somewhere else.

Regarding automatically installing import hook, I was abit worried that it could cause some weird heisenbugs, because DSDLs don't have any common prefix/namespace to be isolated from other packages. However, since it's the last in the import chain, maybe it's not a big deal? 

> If I remember our conversation correctly, there will be no default value for `CYPHALPATH` (i.e., empty), but for `PYCYPHALPATH` the default would be either `~/.pycyphal` (or some equivalent on non-GNU platforms) or `$CYPHALPATH[0]/.pycyphal`. Which one should we go with?

I suggest `~/.pycyphal`, because `$CYPHALPATH[0]/.pycyphal` is not really intuitive without knowing internals and users might get confused if behavior changes on reordering paths.

pavel-kirienko:
>Yes, although might I suggest to use `CYPHAL_PATH` and `PYCYPHAL_PATH`?

Uhm it's fine, but I thought that PATH variables usually come without the underscore? Either way is fine.

>Regarding automatically installing import hook, I was abit worried that it could cause some weird heisenbugs, because DSDLs don't have any common prefix/namespace to be isolated from other packages. However, since it's the last in the import handler chain, maybe it's not a big deal?

I think it should be installed by default but we may want a function for the removal of the import hook for those special cases where the default behavior is undesirable. I am strongly in favor of having the hook installed by default with an explicit opt-out because most applications will want this default behavior.

>I suggest `~/.pycyphal`, because `$CYPHALPATH[0]/.pycyphal` is not really intuitive without knowing internals and users might get confused if behavior changes on reordering paths.

Okay.

----

If you're free at 19:00 our time today and would like to discuss this briefly, you would be welcome to attend our resurrected weekly call. See details here: https://forum.opencyphal.org/t/weekly-dev-call/173

chemicstry:
> I think it should be installed by default but we may want a function for the removal of the import hook for those special cases where the default behavior is undesirable. I am strongly in favor of having the hook installed by default with an explicit opt-out because most applications will want this default behavior.

I will try to implement removal then.

> If you're free at 19:00 our time today and would like to discuss this briefly, you would be welcome to attend our resurrected weekly call. See details here: https://forum.opencyphal.org/t/weekly-dev-call/173

Sorry, I won't be able to join.

chemicstry:
Removal of the import hook is a bit complicated because it is inserted into a global `sys.meta_path` variable, which can be modified by other packages, so you can't rely on index to remove later. You could iterate it and remove by typename, but I think it is far cleaner to add an environment variable, which disables the insertion of default import hook. Which is what I've done.

pavel-kirienko:
image

chemicstry:
:D I mean if you think it would be better to add a remove function, then I can do that instead. However, I see it way more complicated. For example if you use multiple import hooks with different configurations then you also need to differentiate between them. Maybe generating a unique id, which is returned as a handle by `install_import_hook` and can be used for removal later.

Just to be clear, the way it works now is installed by default, but can be opted out with env variable.

I'm trying to fix CI now and this should be ready to go if you don't have any other comments?

chemicstry:
I'm not sure I understand the CI error `TypeError: 'type' object is not subscriptable`, seems to be complaining about `Optional[bool]`. I tried running `nox --non-interactive --session demo check_style docs` locally and it all passes, but not on CI.

demo/demo_app.py
```diff
@@ -7,33 +7,19 @@
 import pathlib
 import asyncio
 import logging
-import importlib
+
```
pavel-kirienko, line 10: Oops

demo/demo_app.py
```diff
@@ -7,33 +7,19 @@
...
 import pycyphal
 
 # Production applications are recommended to compile their DSDL namespaces as part of the build process. The enclosed
 # file "setup.py" provides an example of how to do that. The output path we specify here shall match that of "setup.py".
-# Here we compile DSDL just-in-time to demonstrate an alternative.
+# However, if compiled modules are not found, the pycyphal import hook will compile them from sources defined in CYPHAL_PATH.
 compiled_dsdl_dir = pathlib.Path(__file__).resolve().parent / ".demo_dsdl_compiled"
 
 # Make the compilation outputs importable. Let your IDE index this directory as sources to enable code completion.
 sys.path.insert(0, str(compiled_dsdl_dir))
```
pavel-kirienko, line 20:
```suggestion
```

I don't suppose these are necessary anymore, right? Let's keep the demo clean.

pycyphal/dsdl/_import_hook.py
```diff
@@ -0,0 +1,136 @@
...
+
+        compiled_module_dir = pathlib.Path(self.output_directory, root_namespace)
+        module_location = compiled_module_dir.joinpath("__init__.py")
+        submodule_locations = [str(compiled_module_dir)]
+
+        return spec_from_file_location(fullname, module_location, submodule_search_locations=submodule_locations)
+
+
+def get_default_lookup_dirs():
+    return os.environ.get("CYPHAL_PATH", "").split(os.pathsep)
```
pavel-kirienko, line 88: This will fail on Windows because you're using `:` in the orc file but it would expect `;`. Maybe we should normalize the value first?

demo/launch.orc.yaml
```diff
@@ -2,6 +2,8 @@
 # Read the docs about the orc-file syntax: yakut orchestrate --help
 
 # Shared environment variables for all nodes/processes (can be overridden or selectively removed in local scopes).
+CYPHAL_PATH: "./public_regulated_data_types:./custom_data_types"
+PYCYPHAL_PATH: ".pycyphal_generated"
 YAKUT_COMPILE_OUTPUT: .yakut_compiled
 YAKUT_PATH: .yakut_compiled
 # Here we use Yakut for compiling DSDL. Normally one should use Nunavut though: https://github.com/OpenCyphal/nunavut
```
pavel-kirienko, line 9: This PYTHONPATH is no longer needed.

demo/run.sh
```diff
@@ -0,0 +1,6 @@
+export UAVCAN__SUB__TEMPERATURE_SETPOINT__ID=2345
```
pavel-kirienko, line 1: This file should not be here, you can just run the orc file to the same effect.

pycyphal/dsdl/_import_hook.py
```diff
@@ -0,0 +1,136 @@
...
+        self.lookup_directories = list(map(str, lookup_directories))
+        self.output_directory = output_directory
+        self.allow_unregulated_fixed_port_id = allow_unregulated_fixed_port_id
+        self.root_namespace_directories: Iterable[pathlib.Path] = list()
+
+        # Build a list of root namespace directories from lookup directories.
+        # Any dir inside any of the lookup directories is considered a root namespace.
+        for dir in self.lookup_directories:
+            for namespace in pathlib.Path(dir).iterdir():
+                if namespace.is_dir():
```
pavel-kirienko, line 40: I think we should also check whether it matches the namespace name regex to avoid the inclusion of `.git` et al.

pycyphal/dsdl/_import_hook.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        Returns true if given root namespace exists in output directory (compiled).
+        """
+        return pathlib.Path(self.output_directory, root_namespace).exists()
+
+    def find_spec(self, fullname, path, target=None):
+        _logger.debug("Attempting to load module %s as DSDL", fullname)
+
+        # DSDL root namespace is equal to module name
+        root_namespace = fullname
```
pavel-kirienko, line 63: If my namespace is named `if` the module name would be `if_`, this would break. [`pycyphal.dsdl.get_attribute()`](https://github.com/OpenCyphal/pycyphal/blob/338e206b375c230b96411ca4d5249087273bce36/pycyphal/dsdl/_composite_object.py#L146-L158) has a retry to handle this.

pavel-kirienko (changes requested):

demo/demo_app.py
chemicstry: These demonstrate the usage of `setup.py` where types are precompiled for distribution instead of compiling at runtime.

pavel-kirienko:
>I mean if you think it would be better to add a remove function, then I can do that instead. However, I see it way more complicated.

It's fine. I mostly care about the default case.

>I'm not sure I understand the CI error TypeError: 'type' object is not subscriptable, seems to be complaining about Optional[bool]. I tried running nox --non-interactive --session demo check_style docs locally and it all passes, but not on CI.

Seems like the stack trace is pointing to the wrong line. The error is two lines above where it says this:

```
lookup_directories: Optional[list[_AnyPath]] = None,
```

Replace `list` with `Sequence` from typing.

It passes on your machine because you are using a more recent version of Python.

demo/demo_app.py
pavel-kirienko: Eh, then it should be moved to the place where the `setup.py` example is given.

pycyphal/dsdl/_import_hook.py
chemicstry: Oh yeah, I was wondering how to handle this. Ideally env vars should have OS-specific path separator, but it's hard to do for static configuration files like orc. We could split the path by both `:` and `;` if there are no other OS-specific path separators?

pavel-kirienko: How about this:

```suggestion
    return os.environ.get("CYPHAL_PATH", "").replace(os.pathsep, ";").split(";")
```

demo/run.sh
chemicstry: For some reason `yakut orc` crashes msys bash terminal on windows, but I didn't have time to investigate. I will remove the file

pavel-kirienko: Sorry, the orchestrator is incompatible with Windows indeed :D I thought you are on GNU/Linux.

chemicstry: It uses bash commands so it should work in msys terminal (git bash), but something goes really wrong and entire terminal crashes 🤷

pycyphal/dsdl/_import_hook.py
chemicstry: I assume it would be best to use `check_name` from `pydsdl`, but it is private. Is there a regex for a valid namespace somewhere?

pavel-kirienko: Yes, section 3.2.2:

image

`[a-zA-Z_][a-zA-Z0-9_]*`

pycyphal/dsdl/_import_hook.py
```diff
@@ -0,0 +1,149 @@
...
+        return spec_from_file_location(fullname, module_location, submodule_search_locations=submodule_locations)
+
+
+def get_default_lookup_dirs():
+    return os.environ.get("CYPHAL_PATH", "").replace(os.pathsep, ";").split(";")
+
+
+def get_default_output_dir():
+    pycyphal_dir = os.environ.get("PYCYPHAL_PATH", pathlib.Path.home().joinpath(".pycyphal"))
+    return pathlib.Path(pycyphal_dir, "compiled")
```
pavel-kirienko, line 106: Why do we need this one extra directory here? Why not simply `~/.pycyphal`? As a user, I would expect a PATH directory to point directly to the resource of interest.

pavel-kirienko (changes requested):
- Please merge the master branch and resolve the conflict. The new version number should be v1.10.

- Please drop a note about this feature to CHANGELOG.rst.

pavel-kirienko:
Also, can you please update this section: https://pycyphal.readthedocs.io/en/stable/pages/architecture.html#dsdl-support

The source file is here: https://github.com/OpenCyphal/pycyphal/blob/master/docs/pages/architecture.rst#dsdl-support

pycyphal/dsdl/_import_hook.py
chemicstry: My thinking was that `PYCYPHAL_PATH` refers to general directory containing everything related to pycyphal. Maybe in the future it could contain some config files, cache, register database, whatever. If we don't include additional `compiled` directory then it becomes ambiguous what is what. So it's kind of future insurance.

If directory is to only compiled DSDL's then maybe env var should be `PYCYPHAL_COMPILED_PATH`? Anyway, let me know what you think.

pavel-kirienko: >Maybe in the future it could contain some config files, cache, register database, whatever.

I think that would necessitate a separate variable that doesn't need to be called `*_PATH`. What I think is crucial is that `PYCYPHAL_PATH` points directly to the directory where the compiled modules are located, otherwise it's a usability disaster.

pavel-kirienko:
The CI says (in a very roundabout way) that the demo app is failing to start. Maybe there's an issue with import hooks or something.

chemicstry:
Oops, the commits linked to issue, not intended. I'm still trying to fix the CI, but we can squash the commits afterwards.

pavel-kirienko:
@chemicstry FYI, your progress can be accelerated significantly if you run Nox locally. There are some instructions available in the contributor's guide. Just running `nox -s test` is all it takes (IIRC).

chemicstry:
> @chemicstry FYI, your progress can be accelerated significantly if you run Nox locally. There are some instructions available in the contributor's guide. Just running `nox -s test` is all it takes (IIRC).

I've used other nox commands (`check_style`, `docs`), but `test` doesn't work on Windows, at very least it needs socketcan.

pavel-kirienko:
Nox test works on Windows in the CI environment, no SocketCAN needed ¯\\\_(ツ)\_/¯

https://github.com/OpenCyphal/pycyphal/blob/65d48545c65c3b79c01cb427dd5e083d51426eb7/.appveyor.yml#L67

Seems like there's one remaining issue with home directory detection and then we should be good to merge this.

chemicstry:
> Nox test works on Windows in the CI environment, no SocketCAN needed ¯\_(ツ)_/¯

🤦 that flew over my head. After some hiccups managed to run tests on Windows. However, you should update `ncat.exe` as Windows 11 thinks it's a virus. The latest release from https://nmap.org/download#windows works fine (although needs extra bundled DLLs), but I felt that you would like to update binary files yourself instead of a random stranger.

pavel-kirienko (approved):
@chemicstry are we good to merge now?

pavel-kirienko:
>you should update ncat.exe as Windows 11 thinks it's a virus

The default security policy of Windows 11 where it would randomly delete your files is also somewhat reminiscent of a virus
(╯°□°)╯︵ ┻━┻

chemicstry:
> @chemicstry are we good to merge now?

Yes, I think it's ready now, finally...

PR #238, opened 2022-08-17
Remove VERSION and add version.py

silverv:
This is necessary pycyphal to be installable using pyinstaller.


pycyphal/__init__.py
```diff
@@ -32,10 +32,10 @@
 """
 
 import os as _os
-from importlib.resources import read_text as _read_text
 
 
-__version__ = _read_text(__name__, "VERSION", encoding="utf8").strip()
+from version import __version__ as _v
+__version__ = _v
```
pavel-kirienko, line 38: Can't you just say `from version import __version__ as __version__`?

pavel-kirienko (changes requested):
I think `version.py` should be `_version.py` because it is not part of the API.

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [metadata]
 name            = pycyphal
-version         = file: pycyphal/VERSION
+version         = attr: pycyphal.__version__
```
pavel-kirienko, line 3: This import will fail if the library dependencies (like NumPy) are not installed. Did you mean to say `pycyphal._version.__version__`?

pavel-kirienko (changes requested):

silverv:
As discussed we should look at the CI green of the previous commit and consider the changes made inbetween trivial. Maybe running tests locally to save 4 hours.

pavel-kirienko (changes requested):
Please run black (the code is misformatted) and leave CI running (do not cancel it).

silverv:
It's not misformatted and I will not cancel the CI anymore

pavel-kirienko (approved):

PR #260, opened 2022-11-18
Add ability to spoof media-layer frames

silverv:


pavel-kirienko (changes requested):
Unit tests are missing.

pycyphal/transport/can/_can.py
```diff
@@ -253,6 +254,18 @@ def make_tracer() -> CANTracer:
         """
         return CANTracer()
 
+    async def spoof_frames(self, frames: typing.Sequence[DataFrame]) -> None:
+        async with self._media_lock:
+            if self._maybe_media is None:
+                raise pycyphal.transport.ResourceClosedError(f"{self} is closed")
+            await self._maybe_media.send(
+                [Envelope(f, loopback=False) for f in frames],
+                monotonic_deadline=asyncio.get_running_loop().time() + 1.0,
```
pavel-kirienko, line 263: Do we not want to expose the transmission deadline as an argument for this method? I think it would make the logic in Yukon more explicit.

setup.cfg
```diff
@@ -123,29 +123,37 @@ implicit_reexport      = False
 mypy_path =
     .compiled
 
+
```
pavel-kirienko, line 126: uh oh

tests/transport/can/_can.py
```diff
@@ -1187,6 +1186,93 @@ def on_peep(args: typing.Sequence[typing.Tuple[Timestamp, can.media.Envelope]])
...
+            format=can.media.FrameFormat.EXTENDED,
+            identifier=0x123,
+            data=bytearray(b"124"),
+        ),
+    ]
+    frames_confirmed_received = {str(frame.data): False for frame in list_of_frames}
+    media1_confirmed_received = {str(frame.data): False for frame in list_of_frames}
+    media2_confirmed_received = {str(frame.data): False for frame in list_of_frames}
+    all_frames_received_event = asyncio.Event()
+    all_media1_received_event = asyncio.Event()
```
pavel-kirienko, line 1225: The media1 reception pipeline is not used anywhere and should be removed.

silverv:
I have since gotten distracted with fixing Pycyphal itself and I should come back to the real issue that this pull request is actually done and we should merge it.

silverv:
But that comes after I revert all the fixes that didn't manage to fix Pycyphal and didn't manage to make the life of a developer easier.

pavel-kirienko:
The MyPy problem was caused by a regression in it: https://github.com/python/mypy/issues/13627. I fixed it by upgrading to the latest MyPy; this necessitated some minor additional changes throughout (see the latest commit).

There was a type error and two logical errors in spoof_frames which I fixed. The unit test needs to be updated per my request above and also the `own` flag needs to be checked in the test.

pavel-kirienko (approved):

PR #265, opened 2022-12-09
Allow more pythoncan configuration options

### LINKED ISSUE #264: Support more python-can configuration options
Current [implementation](https://github.com/OpenCyphal/pycyphal/blob/master/pycyphal/transport/can/media/pythoncan/_pythoncan.py#L168) splits environment variable by colon and checks if there are excatctly 2 parts. This does not work with some of the python-can configuration options, like SLCAN over TCP where channel name is `socket://192.168.254.254:5000` instead of just `COM1`.

Would it make sense to split just the first colon and ignore the rest? Or would it be better to use some other format than colon?

chemicstry:
Fixes #264


pavel-kirienko:
One test needs to be updated, please check the CI output.

pavel-kirienko (approved):

chemicstry:
> One test needs to be updated, please check the CI output.

Sorry about that, running tests on windows is a new set of problems on each machine...

One test failed, but it looks unrelated to me. Could you check and/or rerun CI?

PR #271, opened 2023-01-29
Make NodeTracker's GetInfo request priority configurable

Finwood:
Previously, the `pycyphal.application.node_tracker.NodeTracker` always sent out GetInfo requests with the fixed `OPTIONAL` priority. While this seems like a sensible default, there are situations where a higher priority may be desirable.

This PR adds the `NodeTracker.get_info_priority` property, by which the user may change this priority. The default behavior remains unchanged.


pavel-kirienko (approved):
Nice. Could you please also bump the minor version and add a line to the changelog? Thanks!

PR #253, opened 2022-10-09
[WIP] Moving Cyphal/UDP to multicast

maksimdrachov:
This MR is based on [this](https://forum.opencyphal.org/t/cyphal-udp-architectural-issues-caused-by-the-dependency-between-the-nodes-ip-address-and-its-identity/1765) proposol discussed in the OpenCyphal forum.

In short, the changes can be broken down into three pieces:

### 1. Datagram header format

*Current*:

```
uint8 version = 0
uint8 priority
void16 
uint32 frame_index_eot
uint64 transfer_id
void64
```

*Proposal*:

```
uint8 version = 1   # UPDATE
uint8 priority
uint16 source_node_id   # NEW
uint32 frame_index_eot
uint64 transfer_id
void64
```

Note: `version` will be bumped to 1, however no backward-compatibility changes are made (since the protocol is still in development).

### 2. Message

*Current*:

```
    fixed         reserved
   (9 bits)       (3 bits)
   ________          _
  /        \        / \
  11101111.0ddddddd.000sssss.ssssssss
  \__/      \_____/    \____________/
(4 bits)    (7 bits)      (13 bits)
  IPv4      subnet-ID     subject-ID
multicast   \_______________________/
 prefix             (23 bits)
            collision-free multicast
               addressing limit of
              Ethernet MAC for IPv4
```

*Proposal*:

```
    fixed   message  reserved
   (9 bits) select.  (3 bits)
   ________   res.|  _
  /        \     vv / \
  11101111.0ddddd00.000sssss.ssssssss
  \__/      \___/      \____________/
(4 bits)   (5 bits)       (13 bits)
  IPv4     subnet-ID      subject-ID
multicast   \_______________________/
 prefix             (23 bits)
            collision-free multicast
               addressing limit of
              Ethernet MAC for IPv4
```

### 3. Service

*Current*: regular unicast

*Proposal*:

```
    fixed          service
   (9 bits)  res.  selector
   ________      ||
  /        \     vv
  11101111.0ddddd01.nnnnnnnn.nnnnnnnn
  \__/      \___/   \_______________/
(4 bits)   (5 bits)     (16 bits)
  IPv4     subnet-ID     node-ID
multicast   \_______________________/
 prefix             (23 bits)
            collision-free multicast
               addressing limit of
              Ethernet MAC for IPv4
```

### TODO

- [x] Datagram
- [x] Message
- [x] Service
[...]

PR #276, opened 2023-03-07
Demo: Fix the race condition during concurrent DSDL compilation from multiple processes

### LINKED ISSUE #274: Remove the usage of "yakut compile" from the demo along with the related env vars
https://github.com/OpenCyphal/yakut/pull/77

pavel-kirienko:
Fixes #274 

@maksimdrachov this should resolve the Python 3.8 failure in your branch. I am planning to merge it into master so that you can merge it into your branch later.


pavel-kirienko:
The 3.10 test is broken because of a deficiency of the test fixture that is already addressed in #253, which is to be merged next.

PR #280, opened 2023-03-15
Added UDPTransportStatistics

maksimdrachov:
To close: https://github.com/OpenCyphal/pycyphal/issues/279

> Create a new zero-initialized/empty entry in the `received_datagrams` dict when a new session is created using self.`_statistics.received_datagram.setdefault(...)` and pass a reference to it to the session constructor. Let the session update the counters while it exists. When the session is terminated, the corresponding entry in the `received_datagram` dict is not to be removed.


pavel-kirienko (approved):
Please bump the patch version.

pavel-kirienko:
See https://github.com/OpenCyphal/pycyphal/issues/281

PR #283, opened 2023-03-19
Updated the serial header format

maksimdrachov:
https://github.com/OpenCyphal/pycyphal/issues/266


pycyphal/transport/serial/_frame.py
```diff
@@ -132,58 +157,75 @@ def parse_from_cobs_image(image: memoryview) -> typing.Optional[SerialFrame]:
...
-                data_specifier = pycyphal.transport.MessageDataSpecifier(int_data_spec)
-            else:
-                if int_data_spec & (1 << 14):
-                    role = pycyphal.transport.ServiceDataSpecifier.Role.RESPONSE
+        try:
+            if version == SerialFrame.VERSION:
+                crc = pycyphal.transport.commons.crc.CRC16CCITT()
+                crc.add(image[: _HEADER_FORMAT_NO_CRC.size])
+                if header_crc != crc.value:
+                    print("header CRC mismatch")
```
pavel-kirienko, line 184: image

pavel-kirienko (changes requested):

pycyphal/transport/serial/_frame.py
```diff
@@ -54,41 +64,56 @@ def __post_init__(self) -> None:
...
+            0,  # user_data
         )
-        header += pycyphal.transport.commons.crc.CRC32C.new(header).value_as_bytes
-        assert len(header) == _HEADER_SIZE
 
-        out_buffer[0] = self.FRAME_DELIMITER_BYTE
+        crc = pycyphal.transport.commons.crc.CRC16CCITT()
+        crc.add(header_memory)
+        header_crc = crc.value
+        header = header_memory + header_crc.to_bytes(2, "little")
```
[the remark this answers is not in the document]
maksimdrachov: Yes, got stuck on this for a bit figuring out why it was failing

pycyphal/transport/serial/_frame.py
maksimdrachov: image

maksimdrachov: I'll see what i can do

pavel-kirienko: I rather meant why is there a `print()` call...

maksimdrachov: ah oops

pycyphal/transport/serial/_frame.py
```diff
@@ -132,58 +155,73 @@ def parse_from_cobs_image(image: memoryview) -> typing.Optional[SerialFrame]:
...
+                version,
+                int_priority,
+                source_node_id,
+                destination_node_id,
+                data_specifier_snm,
+                transfer_id,
+                frame_index_eot,
+                user_data,
+                _,  # header CRC
+            ) = _HEADER_FORMAT.unpack_from(image)
```
pavel-kirienko, line 173: I think you should be using `_HEADER_FORMAT_NO_CRC` here, and `_HEADER_FORMAT` should be removed completely as it is no longer needed. Same goes for the UDP transport.

pavel-kirienko (changes requested):

pycyphal/transport/serial/_frame.py
```diff
@@ -132,58 +154,72 @@ def parse_from_cobs_image(image: memoryview) -> typing.Optional[SerialFrame]:
...
-
-        try:  # https://github.com/OpenCyphal/pycyphal/issues/176
-            data_specifier: pycyphal.transport.DataSpecifier
-            if int_data_spec & (1 << 15) == 0:
-                data_specifier = pycyphal.transport.MessageDataSpecifier(int_data_spec)
-            else:
-                if int_data_spec & (1 << 14):
-                    role = pycyphal.transport.ServiceDataSpecifier.Role.RESPONSE
+        try:
+            if version == SerialFrame.VERSION:
```
pavel-kirienko, line 176: This is fine, but I think the commonly accepted practice is to use early return so that you don't have to indent the conditional block so much, as it hurts readability (especially in a language that follows the off-side rule).

pavel-kirienko (approved):

PR #285, opened 2023-03-22
Fix CYPHAL_PATH issue

maksimdrachov:
See comments in https://github.com/OpenCyphal/pycyphal/issues/278

### DIFF
--- a/pycyphal/dsdl/_import_hook.py
+++ b/pycyphal/dsdl/_import_hook.py
@@ -101,7 +101,9 @@ class DsdlMetaFinder(MetaPathFinder):
 
 
 def get_default_lookup_dirs() -> Sequence[str]:
-    return os.environ.get("CYPHAL_PATH", "").replace(os.pathsep, ";").split(";")
+    dirs = os.environ.get("CYPHAL_PATH", "").replace(os.pathsep, ";").split(";")
+    dirs = [dir for dir in dirs if dir]  # filter out empty strings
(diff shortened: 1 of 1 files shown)


pycyphal/dsdl/_import_hook.py
```diff
@@ -101,7 +101,9 @@ def find_spec(
 
 
 def get_default_lookup_dirs() -> Sequence[str]:
-    return os.environ.get("CYPHAL_PATH", "").replace(os.pathsep, ";").split(";")
+    dirs = os.environ.get("CYPHAL_PATH", "").replace(os.pathsep, ";").split(";")
+    dirs = [dir for dir in dirs if dir]  # filter out empty strings
```
pavel-kirienko, line 105: 1. Add `.strip()` to also ignore whitespace-only strings.
2. `dir` is a [built-in entity](https://docs.python.org/3/library/functions.html#dir) in Python and the linter will (or should) complain about this.

```suggestion
    dirs = [d for d in dirs if d.strip()]  # filter out empty strings
```

pavel-kirienko (changes requested):
Please also bump the patch number.

pavel-kirienko (approved):

pavel-kirienko:
future note: please do not close issues until the pull request fixing them is actually merged (github will do it for you automatically anyway)

maksimdrachov:
noted!

PR #295, opened 2023-04-27
Fix sniffer test

maksimdrachov:
Tried setting the priority, as per: https://pycyphal--293.org.readthedocs.build/en/293/pages/faq.html

It doesn't fix the issue.

image

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -65,6 +65,13 @@ def _configure_host_environment() -> None:
         ctypes.WinDLL("NTDLL.DLL").NtSetTimerResolution(5000, 1, ctypes.byref(t))
         _logger.info("System timer resolution: %.3f ms", t.value / 10e3)
 
+    if sys.platform.startswith("darwin") or sys.platform.startswith("linux"):
+        import psutil
+        import os
+
+        p = psutil.Process(os.getpid())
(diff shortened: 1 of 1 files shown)


silverv:
I also observed that even a realtime priority on Windows lead to the timestamp being off

tests/conftest.py
```diff
@@ -65,6 +65,13 @@ def execute(*cmd: typing.Any, ensure_success: bool = True) -> typing.Tuple[int,
         ctypes.WinDLL("NTDLL.DLL").NtSetTimerResolution(5000, 1, ctypes.byref(t))
         _logger.info("System timer resolution: %.3f ms", t.value / 10e3)
 
+    if sys.platform.startswith("darwin") or sys.platform.startswith("linux"):
+        import psutil
+        import os
+
+        p = psutil.Process(os.getpid())
+        p.nice(-20)
```
pavel-kirienko, line 73: For this to work on GNU/Linux, the ulimit on nice needs to be lifted first; this can be done through the noxfile https://access.redhat.com/solutions/61334

Although at this point we should suspect that the failures on macOS are not performance-related and require further investigation.

pavel-kirienko:
Please resubmit if this is still relevant.

PR #305, opened 2023-07-10
Socketcand implimentation

wiboticanders:
Added implimention of the socketcand interface. The socketcand interfaces requires additional variables 'host' and 'port' so that it can communicate w a remote computer. This adds those variables as optional, and allows for a socketcand constructor. Socketcand is already functional with python-can, here is the library: https://github.com/linux-can/socketcand. I was not totally sure how you guys preferred this to be implemented, so I have a second implementation that does not add new init vars, and instead adds the new variables to the existing ifacename variable: https://github.com/wiboticanders/pycyphal/blob/noNewInitVars/pycyphal/transport/can/media/pythoncan/_pythoncan.py.

### DIFF
--- a/pycyphal/transport/can/media/pythoncan/_pythoncan.py
+++ b/pycyphal/transport/can/media/pythoncan/_pythoncan.py
@@ -79,6 +79,8 @@ class PythonCANMedia(Media):
         iface_name: str,
         bitrate: typing.Union[int, typing.Tuple[int, int]],
         mtu: typing.Optional[int] = None,
+        host: typing.Optional[str] = None,
+        port: typing.Optional[int] = None,
         *,
         loop: typing.Optional[asyncio.AbstractEventLoop] = None,
     ) -> None:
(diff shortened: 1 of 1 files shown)


pavel-kirienko:
Hi @wiboticanders, thank you for the pull request, this is a useful feature. I've been mostly using [cannelloni](https://forum.opencyphal.org/t/connecting-to-a-can-bus-via-a-remote-computer/1726) but socketcand is a better tool for many use cases.

I don't like the way it integrates, though. The fact that you had to add multiple special cases to support this interface suggests that either the original PythonCAN wrapper design is unnecessarily rigid, or the new interface doesn't fit the model used in the existing wrapper implementation. I think it is the latter; hence, I think the socketcand support should be split into a separate media layer class, `SocketcandMedia`. Initially, it can be implemented using PythonCAN, but eventually, we should like to define a native implementation because PythonCAN is not the best library out there, and it is desirable to reduce the dependency on it.

Once SocketcandMedia is implemented, we will want to add a handler for it to the `transport_factory`:

https://github.com/OpenCyphal/pycyphal/blob/01b9a9b57143bc916bfa2fa26102a787b4e558e9/pycyphal/application/_transport_factory.py#L290-L303

PR #306, opened 2023-07-11
SocketcandMedia Media class added for socketcand functionality

wiboticanders:
Although this runs on PythonCAN, changing the PythonCANMedia class to function with socketcand is not very clean because socketcand has 2 additional variables compared to all the other PythonCAN compatible interfaces on pycyphal.
Socketcand can be referenced here: https://github.com/linux-can/socketcand , it is used for remote CAN access of udp socket. As far as copyright goes, because this is very similar to the PythonCANMedia class and I reused code from it I believe that the copyright still applies.

PR #308, opened 2023-07-12
Updated version of the socketcandMedia class.

wiboticanders:
This fixes the unnecessary pieces left from the pythonCANMedia class, such as the FD vs Classic dataclasses, Can Bus Options, and the constructor class. It also updates the way parameters are inputed to be all in the iface_name param like 'socketcand:can:host:port'.


pavel-kirienko:
Is this meant to replace https://github.com/OpenCyphal/pycyphal/pull/306? Please do not create new pull requests for the same change to avoid losing context.

wiboticanders:
Apologies, accident. #306 is updated.

PR #319, opened 2024-01-03
Improve demo documentation

maksimdrachov:


pavel-kirienko (changes requested):
See comrade, `yakut compile` is not needed anymore. `YAKUT_PATH` went the same way. The command itself will be removed from Yakut during the next sprint. All you need to do is set `CYPHAL_PATH` pointing to where your DSDL root directories are, and compilation will happen automatically as needed. You will update the docs accordingly yes?

maksimdrachov:
> next sprint

so, give or take, somewhere in the next decade?

pavel-kirienko:
My bad, I meant the next five-year plan, aka *pyatiletka*.

pavel-kirienko (approved):

Issue #323, opened 2024-01-12
Cyphal/UDP: the default value of local_node_id should be absent, not zero


pavel-kirienko:
https://github.com/OpenCyphal/pycyphal/blob/088932f6690212b1c90124624352e3b4d15ae495/pycyphal/transport/udp/_udp.py#L67

PR #337, opened 2024-05-03
Cyphal/UDP: remove the default local_node_id

songmeo:
addresses #323


pycyphal/transport/udp/_udp.py
```diff
@@ -64,7 +64,7 @@ class UDPTransport(pycyphal.transport.Transport):
     def __init__(
         self,
         local_ip_address: IPAddress | str,
-        local_node_id: typing.Optional[int] = 0,
+        local_node_id: typing.Optional[int] = None,
```
pavel-kirienko, line 67:
```suggestion
        local_node_id: typing.Optional[int],
```

I originally wrote `None` in the issue but then realized that the default should be absent for consistency with the other transports, and updated the issue accordingly.

pavel-kirienko (changes requested):
Please bump the patch version.

pycyphal/transport/udp/_udp.py
pavel-kirienko: Yes please.

pavel-kirienko:
Please also fix the failed test affected by this change

pavel-kirienko:
(history rewrite is not necessary because all commits will be squashed upon merging)

pavel-kirienko:
This changeset has been merged into `dev`; I will merge it into master next.

PR #341, opened 2024-06-19
Fix socketcan timestamp on newer 32 bit kernel

tomasjakubik:
Kernel now defines old and new SO_TIMESTAMP and can send both of them in ancillary data.


pycyphal/_version.py
```diff
@@ -1 +1 @@
-__version__ = "1.19.0"
+__version__ = "1.20.0"
```
pavel-kirienko, line 1: Sorry, since this is a fix, can you please bump the [patch](https://semver.org/) version instead?

pavel-kirienko (changes requested):
I'm not sure why the workflow was skipped either. GitHub Actions are being weird lately.

pavel-kirienko (approved):

PR #347, opened 2025-04-23
fix: add base class to exception

hannesweisbach:
pycyphal.application.file.RemoteFileError was introduced as common exception base class/tag type to catch exceptions resulting from remote file operations using the Cyphal remote file system interface.

In Python exceptions must inherit from BaseException, otherwise a TypeError will occur:

TypeError: exceptions must derive from BaseException

RemoteFileError does not inherit from BaseException and can't thus be used in except clauses.

This fix lets RemoteFileError inherit from RuntimeError, so that RemoteFileError can be used as intended.

this fixes commit 40e14821285f48ad997cb3f7a8457d79a88db082


hannesweisbach:
Hm, I'm not sure if this is a major change. If I consider `RemoteFileError` as part of the API, then this would require a major version, no?

pavel-kirienko:
It's worthy of a patch version bump only, since this is a fix that doesn't affect backward compatibility of the API

pavel-kirienko (approved):

pavel-kirienko:
Thank you! I have moved your commit to the `dev` branch and I will merge it next.

Issue #270, opened 2023-01-23
Needs remove_import_hooks


silverv:
It should do something like this:
```
for meta_path in sys.meta_path.copy():
    if isinstance(meta_path, DsdlMetaFinder):
        sys.meta_path.remove(meta_path)
```
It should be available in the `__init__.py` file in the pycyphal.dsdl module

pavel-kirienko:
Also, `install_import_hook` needs to be renamed into `add_import_hook` and the old name should be deprecated. The docs need to make it clear that the function does not modify the existing import hooks.

PR #356, opened 2025-06-04
Add a function for removing import hooks

songmeo:
address #270

### DIFF
--- a/pycyphal/dsdl/__init__.py
+++ b/pycyphal/dsdl/__init__.py
@@ -48,3 +48,14 @@ def generate_package(*args, **kwargs):  # type: ignore  # pragma: no cover
         DeprecationWarning,
     )
     return compile(*args, **kwargs)
+
+
+def install_import_hook(*args, **kwargs):  # type: ignore  # pragma: no cover
+    """Deprecated alias of :func:`add_import_hook`."""
+    import warnings
+
+    warnings.warn(
+        "pycyphal.dsdl.install_import_hook() is deprecated; use pycyphal.dsdl.add_import_hook() instead.",
+        DeprecationWarning,
+    )
+    return add_import_hook(*args, **kwargs)
--- a/pycyphal/dsdl/_import_hook.py
+++ b/pycyphal/dsdl/_import_hook.py
@@ -163,9 +163,6 @@ def add_import_hook(
 
 
 def remove_import_hooks() -> None:
-    """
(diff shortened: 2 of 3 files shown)


pycyphal/dsdl/__init__.py
```diff
@@ -21,7 +21,8 @@
 from ._compiler import compile_all as compile_all
 from ._compiler import GeneratedPackageInfo as GeneratedPackageInfo
 
-from ._import_hook import install_import_hook as install_import_hook
+from ._import_hook import add_import_hook as add_import_hook
```
pavel-kirienko, line 24: This change breaks the API. The ticket calls for deprecation of the old name, not its removal. Please add an `install_import_hook` wrapper here in this file and make it deprecated, similar to `generate_package`.

pycyphal/dsdl/_import_hook.py
```diff
@@ -162,9 +162,18 @@ def install_import_hook(
     sys.meta_path.append(DsdlMetaFinder(lookup_directories, output_directory, allow_unregulated_fixed_port_id))
 
 
+def remove_import_hooks() -> None:
+    """
+    Removes all import hooks.
+    """
```
pavel-kirienko, line 168:
```suggestion
```

Comments that add no useful information need not be added.

tests/dsdl/_compiler.py
```diff
@@ -1,13 +1,16 @@
 # Copyright (c) 2020 OpenCyphal
 # This software is distributed under the terms of the MIT License.
 # Author: Pavel Kirienko <pavel@opencyphal.org>
-
+import sys
```
pavel-kirienko, line 4:
```suggestion

import sys
```

tests/dsdl/_compiler.py
```diff
@@ -1,13 +1,16 @@
...
 import typing
 import logging
 import pathlib
 import tempfile
 import pytest
 import pycyphal.dsdl
+from pycyphal.dsdl._import_hook import DsdlMetaFinder
```
pavel-kirienko, line 11: Protected name imports better be local. Please move this one into the test function scope.

pavel-kirienko (changes requested):
I think we better stage this on a separate branch to release it later with your other pending changes. Please change the merge target branch to `dev`.

pavel-kirienko (approved):

Issue #335, opened 2024-04-21
Automatic DSDL compilation should lock the directory when compilation is in progress


pavel-kirienko:
This is to properly support the common scenario when multiple scripts are launched concurrently and both attempt to compile the same namespace, which is a common occurrence with orchestration.

PR #357, opened 2025-06-10
335 create lockfile

songmeo:
address #335

### DIFF
--- a/pycyphal/dsdl/_compiler.py
+++ b/pycyphal/dsdl/_compiler.py
@@ -191,7 +191,6 @@ def compile(  # pylint: disable=redefined-builtin
             output_directory=output_directory,
         ) as lockfile:
             if lockfile:
-                # Generate code
                 assert isinstance(output_directory, pathlib.Path)
                 code_generator = nunavut.jinja.DSDLCodeGenerator(
                     namespace=root_ns,
--- a/pycyphal/dsdl/_lockfile.py
+++ b/pycyphal/dsdl/_lockfile.py
@@ -30,8 +30,7 @@ class Locker:
         self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None
     ) -> None:
         if self._lockfile is not None:
-            return self.remove()
-        return None
+            self.remove()
 
     def create(self) -> bool:
         """


pycyphal/dsdl/__init__.py
```diff
@@ -38,6 +38,8 @@
 from ._support_wrappers import to_builtin as to_builtin
 from ._support_wrappers import update_from_builtin as update_from_builtin
 
+from pycyphal.dsdl._lockfile import Locker
```
pavel-kirienko, line 41: The locker is not part of the API so it should not be imported here.

pycyphal/dsdl/_compiler.py
```diff
@@ -143,12 +143,11 @@ def compile(  # pylint: disable=redefined-builtin
         # https://forum.opencyphal.org/t/nestedrootnamespaceerror-in-basic-usage-demo/794
         raise TypeError(f"Lookup directories shall be an iterable of paths, not {type(lookup_directories).__name__}")
 
-    output_directory = pathlib.Path(pathlib.Path.cwd() if output_directory is None else output_directory).resolve()
-
     language_context = nunavut.lang.LanguageContextBuilder().set_target_language("py").create()
 
     root_namespace_name: str = ""
     composite_types: list[pydsdl.CompositeType] = []
+    output_directory = pathlib.Path(pathlib.Path.cwd() if output_directory is None else output_directory).resolve()
```
pavel-kirienko, line 150: Why was this moved?

pycyphal/dsdl/_compiler.py
```diff
@@ -171,14 +170,28 @@ def compile(  # pylint: disable=redefined-builtin
...
+    else:
+        root_ns = nunavut.build_namespace_tree(
+            types=[],
+            root_namespace_dir=str(""),
+            output_dir=str(output_directory),
+            language_context=language_context,
+        )
+
+    lockfile = Locker(
+        root_namespace_name=root_namespace_name if root_namespace_name else "support",
```
pavel-kirienko, line 188: Avoid name collision if there is a namespace named `support`.

```suggestion
        root_namespace_name=root_namespace_name if root_namespace_name else "_support_",
```

pycyphal/dsdl/_compiler.py
```diff
@@ -171,14 +170,28 @@ def compile(  # pylint: disable=redefined-builtin
...
+        output_directory=output_directory,
+    )
+    lockfile_created = lockfile.create()
+    if lockfile_created:
```
pavel-kirienko, line 192: The next step is to introduce a context manager. Currently, while this code is logically correct, it is not exception-safe.

```suggestion
    if lockfile.create():
```

pavel-kirienko (changes requested):
Nice.

pycyphal/dsdl/_lockfile.py
```diff
@@ -0,0 +1,57 @@
...
+    def _lockfile_path(self) -> Path:
+        return self._output_directory / f"{self._root_namespace_name}.lock"
+
+    def __enter__(self) -> bool:
+        return self.create()
+
+    def __exit__(
+        self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None
+    ) -> None:
+        return self.remove()
```
pavel-kirienko, line 28: This will fail if create returned false

pavel-kirienko (changes requested):

pycyphal/dsdl/_compiler.py
```diff
@@ -199,28 +185,36 @@ def compile(  # pylint: disable=redefined-builtin
...
-            str(output_directory),
-            quick_fix,
-        )
+    if root_namespace_name is not None:
+        with Locker(
+            root_namespace_name=root_namespace_name,
+            output_directory=output_directory,
+        ) as lockfile:
+            if lockfile:
+                # Generate code
```
pavel-kirienko, line 194:
```suggestion
```

pavel-kirienko (approved):
Nice. This should also be manually tested with Yakut with `PYCYPHAL_LOGLEVEL=DEBUG`: start two Yakut processes simultaneously without the `uavcan` namespace compiled, and make sure only one of them does the compilation while the other is waiting.

pycyphal/dsdl/_lockfile.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def __enter__(self) -> bool:
+        return self.create()
+
+    def __exit__(
+        self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None
+    ) -> None:
+        if self._lockfile is not None:
+            return self.remove()
+        return None
```
pavel-kirienko, line 34:
```suggestion
        if self._lockfile is not None:
            self.remove()
```

PR #362, opened 2025-07-10
Merging dev to master

### LINKED ISSUE #355: Register _strictify() fails to convert long numerical arrays correctly
Currently, `_strictify` attempts to convert all raw numerical arrays into `Natural64`, `Integer64`, or `Real64`, depending on the signedness and kind of the source array elements:

https://github.com/OpenCyphal/pycyphal/blob/71eb3ae06b51820f0f4ec27e2f1edc48f73fad8e/pycyphal/application/register/_value.py#L352-L355

This only works correctly if the source array contains at most 32 elements. If there are more of them, they don't fit in these large 64-bit types, causing errors like:

```python
  File "fluxgrip_software/venv/lib/python3.12/site-packages/yakut/cmd/register_access/_logic.py", line 129, in _access_one
    val.assign_environment_variable(reg_val_str)
  File "fluxgrip_software/venv/lib/python3.12/site-packages/pycyphal/application/register/_value.py", line 159, in assign_environment_variable
    self.assign(numbers)
  File "fluxgrip_software/venv/lib/python3.12/site-packages/pycyphal/application/register/_value.py", line 130, in assign
    res = _do_convert(self._value, _strictify(source))
                                   ^^^^^^^^^^^^^^^^^^
  File "fluxgrip_software/venv/lib/python3.12/site-packages/pycyphal/application/register/_value.py", line 353, in _strictify
    return _strictify(Natural64(s)) if all(x >= 0 for x in s) else _strictify(Integer64(s))
                                                                              ^^^^^^^^^^^^
  File ".pycyphal/uavcan/primitive/array/Integer64_1_0.py", line 63, in __init__
    raise ValueError(f'value: invalid array length: not {value.size} <= 32')
ValueError: value: invalid array length: not 51 <= 32
```

One way to fix it is to pick the type width based on the number of elements in the source list.

songmeo:
This PR resolves #355 #335 #270


pycyphal/dsdl/_lockfile.py
```diff
@@ -0,0 +1,62 @@
...
+import time
+from io import TextIOWrapper
+from pathlib import Path
+from types import TracebackType
+
+_logger = logging.getLogger(__name__)
+
+
+class Locker:
+
```
pavel-kirienko, line 16: Add a brief doc comment explaining the purpose of the class.

pavel-kirienko (approved):
Please bump the minor version number and update the changelog.

PR #363, opened 2025-07-14
build graphviz 13.x

songmeo:
Build graphviz 13.x from source


.github/workflows/test-and-release.yml
```diff
@@ -50,6 +50,14 @@ jobs:
           python --version
           ip link show
 
+          # Install Graphviz 13.x
+          wget https://gitlab.com/api/v4/projects/4207231/packages/generic/graphviz-releases/13.1.0/ubuntu_24.04_graphviz-13.1.0-cmake.deb
+          sudo apt install ./ubuntu_24.04_graphviz-13.1.0-cmake.deb
```
pavel-kirienko, line 55: IIRC apt doesn't have a stable command-line interface, so in automation scripts you should be using `apt-get` instead.

pavel-kirienko (approved):
Please bump the patch version, otherwise CD will fail with a tag conflict.

PR #365, opened 2025-07-16
fix css for inheritance diagram size

### LINKED ISSUE #321: [docs] Class diagrams are scaled incorrectly
This is not reproducible on my local machine but is visible only on ReadTheDocs:

image

The difference is in the scale transform used in the generated SVG file. This is visible on the ReadTheDocs website:

image

This is on my local machine:

image

The obvious thing is that the graphviz (dot) version is different: the latest version in Ubuntu 22.04 is v2.43, which is reported in the generated SVG file, while on my machine (Manjaro) I have v9.0.

We should either set the scale explicitly or make ReadTheDocs use an up-to-date version of Graphviz. One way to do the latter is to grab the DEB package from the release page at https://gitlab.com/graphviz/graphviz/-/releases and install it using a pre-build job as documented in the [readthedocs docs](https://docs.readthedocs.io/en/stable/config-file/v2.html#build-jobs).

songmeo:
Closes #321 

The problem with the diagram scale coming from the inheritance diagram not applying scaling attributes responsively. 

This PR enables CSS to scale inheritance images to 75%.


songmeo:
Please see https://pycyphal--365.org.readthedocs.build/en/365/index.html for the resulting docs.

pavel-kirienko (approved):

PR #367, opened 2025-07-24
Fix bitrate calculation for PCAN interfaces

Finwood:
The existing [PCAN](https://python-can.readthedocs.io/en/stable/interfaces/pcan.html) interface implementation disregarded the configured bitrate preferences, and used static magic-numbers instead. This PR changes the timing calculation by utilizing [`can.BitTimingFd.from_sample_point()`](https://python-can.readthedocs.io/en/stable/bit_timing.html#can.BitTimingFd.from_sample_point), which leads to a cleaner implementation.

I am unsure regarding the target sample point location: my setup works well with 87.5%, and does _not_ work with 75%, but that might be an edge case (low clock precision). Does anyone have recommendations on sample point location?

### DIFF
--- a/pycyphal/transport/can/media/pythoncan/_pythoncan.py
+++ b/pycyphal/transport/can/media/pythoncan/_pythoncan.py
@@ -496,9 +496,9 @@ def _construct_pcan(parameters: _InterfaceParameters) -> can.ThreadSafeBus:
         timing = can.BitTimingFd.from_sample_point(
             f_clock=80_000_000,  # TODO: 80 MHz is a good choice for high data rates, what about lower ones?
             nom_bitrate=parameters.bitrate[0],
-            nom_sample_point=87.5,
+            nom_sample_point=80,
             data_bitrate=parameters.bitrate[1],
-            data_sample_point=87.5,
+            data_sample_point=80,
         )
         _logger.debug("PCAN timing solution: %s", timing)
         return (


pavel-kirienko:
>I am unsure regarding the target sample point location: my setup works well with 87.5%, and does not work with 75%, but that might be an edge case (low clock precision). Does anyone have recommendations on sample point location?

Could be caused by long wiring, or large uncompensated transceiver delays, if we're talking about CAN FD. This might or might not be relevant: https://forum.opencyphal.org/t/on-can-bus-topology-and-termination/1685

[CANopen FD does not recommend placing the sample point past 80% in the arbitration segment](https://www.can-cia.org/fileadmin/cia/documents/publications/cnlm/march_2018/18-1_p28_recommendation_for_the_canfd_bit-timing_holger_zeltwanger_cia.pdf) (at least in the automotive applications, which presumably applies to any other application where wiring length can exceed ~10 meters)

pycyphal/transport/can/media/pythoncan/_pythoncan.py
```diff
@@ -491,32 +491,21 @@ def _construct_pcan(parameters: _InterfaceParameters) -> can.ThreadSafeBus:
...
-        data_tseg1, data_tseg2, data_sjw = 3, 1, 1
-
-        nom_br = int(f_clock / parameters.bitrate[0] / (nom_tseg1 + nom_tseg2 + nom_sjw))
-        data_br = int(f_clock / parameters.bitrate[1] / (data_tseg1 + data_tseg2 + data_sjw))
-        # TODO: validate the result and see if it is within an acceptable range
 
+        timing = can.BitTimingFd.from_sample_point(
+            f_clock=80_000_000,  # TODO: 80 MHz is a good choice for high data rates, what about lower ones?
+            nom_bitrate=parameters.bitrate[0],
+            nom_sample_point=87.5,
```
pavel-kirienko, line 498: Should we set it to 80% to better conform to CANopen & J1939? Will your setup work in that case?

We should do the same for the data segment I reckon.

pavel-kirienko (approved):
Please bump the patch version in `__version__.py`

pycyphal/transport/can/media/pythoncan/_pythoncan.py
Finwood: My setup does not work with 80%, but that is more of a _me_-problem than a pycyphal issue. Setting the default sample point to 80% seems reasonable, as that should be compatible to most setups. I can work with a local fork until my clock issues are resolved...

PR #369, opened 2025-07-28
Fix CAN FD bitrate switching

Finwood:
Set the BRS bit for pythoncan transports when the interface operates in CAN FD mode.

Fixes https://github.com/OpenCyphal/pycyphal/issues/368


pavel-kirienko (commented):
Please move this commit into your other PR and we will release both simultaneously.

PR #377, opened 2026-03-25
Add option to hook into internal pycyphal errors

LukasLendvorsky:
Currently, when pycyphal encounters an error in one of its background asyncio tasks (e.g. a task receiving messages in the background), it simply calls logger.exception. This is fine for debugging with a human looking at the logs, but insufficient for production environments where you need to react to unexpected problems — there is no way of knowing an exception happened except by reading the logs.

This PR introduces set_internal_error_handler, which can be used to register a global hook into pycyphal. This hook will receive all background exceptions when they happen and can react to them as needed (e.g. by raising a fatal error). When no hook is registered, the behavior is unchanged — exceptions are simply logged.


pycyphal/util/__init__.py
```diff
@@ -6,6 +6,9 @@
 The util package contains various entities that are commonly useful in PyCyphal-based applications.
 """
 
+from pycyphal.util.error_reporting import handle_internal_error as handle_internal_error
```
pavel-kirienko: Why does it need to be exported if it's internal?

pycyphal/util/error_reporting.py
```diff
@@ -0,0 +1,33 @@
...
+    Report an internal error: log it via the provided *logger* and invoke the registered error handler.
+
+    :param logger: The logger to use for ``logger.exception``.
+    :param e: The exception to report.
+    :param msg: Optional context message describing where/why the error occurred.
+    """
+    logger.exception(msg or e)
+    if _error_handler is not None:
+        if msg:
+            e.add_note(msg)
```
pavel-kirienko, line 53: add_note is Python 3.11+ only; we must support 3.10 until October

pycyphal/util/error_reporting.py
```diff
@@ -0,0 +1,33 @@
...
+    Register a callback that will be invoked whenever an internal pycyphal component encounters
+    an exception somewhere in background asyncio tasks.
+
+    This is useful to be notified when something goes wrong while receiving messages in the background etc.
+    
+    """
+    global _error_handler  # noqa: PLW0603
+    _error_handler = handler
+
+def handle_internal_error(logger: logging.Logger, e: Exception, msg: str = "") -> None:
```
pavel-kirienko: Callers should defer string formatting to this handler instead of doing it eagerly because we have a number of `__repr__`/`__str__` calls that are not noexcept. Something like this:

```python
    def handle_internal_error(
        logger: logging.Logger,
        e: Exception,
        msg: str = "",
        *args: object,
    ) -> None:
        if msg:
            logger.error(msg, *args, exc_info=e)
        else:
            logger.error("Unhandled internal error", exc_info=e)
        if _error_handler is not None:
            _error_handler(e) # add note if 3.11+
```

Then the call sites become lazy again:

```python
    handle_internal_error(_logger, ex, "fgsfds %s", self)
```

pavel-kirienko (changes requested):
Thanks! The idea is sound. I left a few remarks.

The patch version should be bumped for us to merge this; see CONTRIBUTING.md.

pycyphal/util/error_reporting.py
LukasLendvorsky: I don't understand what problem does this solve? If I don't explicitly do the formatting here, logger will do it anyway and, any exception from __repr__ is still raised and nothing is logged.
My version here behaves same way, only difference might be when the log message is supressed by the log level. But that doesn't seem probable since this is error level.

pavel-kirienko: This is the problem, if `self.__repr__` throws, you're dead:

```
pycyphal.util.handle_internal_error(_logger, ex, f"{self!r}: Could not initialize the publisher")
```

LukasLendvorsky: I understand that, but what I'm saying that if I leave formatting for the logger you are dead as well, that doesn't solve the problem. This reraises the eception as well:
```
import logging
import pytest

class _BrokenRepr:
    def __repr__(self) -> str:
        raise RuntimeError("repr is broken")

    def __str__(self) -> str:
        raise RuntimeError("str is broken")



def test_xy(caplog: pytest.LogCaptureFixture) -> None:  # pyright: ignore[reportUndefinedVariable]
    logger = logging.getLogger("test.error_reporting.broken_repr")
    obj = _BrokenRepr()
    exc = RuntimeError("fail")

    try:
        with caplog.at_level(logging.ERROR, logger="test.error_reporting.broken_repr"):
            logger.error("processing %r failed", obj)
    except RuntimeError:
        # this gets raised and there is nothing in the log
        pass
    
    assert len(caplog.text) > 0, "Logging should produce some output even if __repr__ is broken"
```

LukasLendvorsky: Maybe better solution is to catch exceptions during formatting and if there is any, just log something like "Exception during formatting message" instead.

pavel-kirienko: That sounds reasonable

pycyphal/util/__init__.py
LukasLendvorsky: Removed.

pycyphal/util/error_reporting.py
LukasLendvorsky: Added check

pycyphal/util/error_reporting.py
LukasLendvorsky: Done, also added test to make sure.

LukasLendvorsky:
Updated - thanks for quick review 👍

pavel-kirienko (approved):

PR #391, opened 2026-06-29
Add file_client and file_server examples

maksimdrachov:


examples/file_client.py
```diff
@@ -0,0 +1,161 @@
...
+
+NAME = f"{Path(__file__).stem}/"  # The trailing separator ensures that a random ID will be added.
+TOPIC = "file/read"
+PATH_MAX_LEN = 2048
+DATA_MAX = 4096
+RESPONSE_TIMEOUT = 30.0
+REQUEST_DELIVERY_TIMEOUT = RESPONSE_TIMEOUT / 2.0
+REQUEST_HEADER_FORMAT = "<QH"
+RESPONSE_HEADER_FORMAT = "<IH"
+RESPONSE_HEADER_SIZE = struct.calcsize(RESPONSE_HEADER_FORMAT)
```
pavel-kirienko: These belong in the dataclasses and they should be private with exposed serialize/deserialize methods. Maybe we should even move them into a shared module to reduce duplication between the examples.

examples/file_client.py
```diff
@@ -0,0 +1,161 @@
...
+_logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class FileReadResponse:
+    error: int
+    data: bytes
+
+
+def _encode_request(file_path: str, read_offset: int) -> bytes:
```
pavel-kirienko: same, broken encapsulation

examples/file_server.py
```diff
@@ -0,0 +1,150 @@
...
+    sub = node.subscribe(TOPIC)
+    tasks: set[asyncio.Task[None]] = set()
+    _logger.info("file server ready on %r via %s", TOPIC, transport)
+    try:
+        async for arrival in sub:
+            request = _decode_request(arrival.message)
+            if request is None:
+                _logger.debug("dropping malformed request of size %d", len(arrival.message))
+                continue
+            task = asyncio.create_task(_serve_request(arrival, request), name=f"file:{arrival.breadcrumb.tag}")
```
pavel-kirienko: I think this example would benefit from doing it sequentially to focus more on Cyphal rather than asyncio.

pavel-kirienko (changes requested):
please bump the dev version number so that it's published

